Issue #8, opened 2016-03-02
Parallelization


trentmwillis:
One of the core features I hoped to introduce with this addon is parallelization of tests. There are lots of considerations that went into this, and it finally resulted in needing support from [Testem](https://github.com/testem/testem).

The feature will leverage the new [multiple test pages in parallel](https://github.com/testem/testem/pull/800) config option from Testem once it is released. So, opening this issue to make sure the updates here happen once that is available.

paulius005:
Is there any help that you would need with this going out? Or should we hang tight until the support from Testem comes?

trentmwillis:
Hang tight. Just got news today that the new Testem feature is released, so I plan to work on this sometime within the next week hopefully.

trentmwillis:
This has been completed and is available in version 0.2.1.

PR #11, opened 2016-03-12
Implement parallelization

trentmwillis:
Addressing #8. I think this is ready to go. Just want to do some more testing with actual projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,11 @@
 
 [![Build Status](https://travis-ci.org/trentmwillis/ember-exam.svg)](https://travis-ci.org/trentmwillis/ember-exam)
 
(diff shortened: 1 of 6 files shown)


stefanpenner:
rather then building N different projects (which itself has some issues) https://github.com/trentmwillis/ember-exam/issues/15 would enable a single build to be run multiple ways

trentmwillis:
@stefanpenner this approach doesn't require N projects; it should work with the current approach as well as #15. The parallelization is achieved by simply modifying the Testem config to run multiple test pages, which can be done off a single build.

I'm going to try and wrap up the remaining tests on this over the weekend and hopefully explore #15 sometime next week (probably to-from EmberConf).

stefanpenner:
Ah, sweet!

Issue #34, opened 2016-06-21
introduces 60ms+ slowdown on ember-cli boot 


stefanpenner:
various requires should likely be made lazily/on-demand

the node module time-require can be of help

trentmwillis:
Will look into this tonight

trentmwillis:
#35 addresses this as much as possible.

PR #35, opened 2016-06-21
Make requires lazy where possible

trentmwillis:
Addressing #34.

Unfortunately, I think the bulk of slowness comes from [requiring the Test Command](https://github.com/trentmwillis/ember-exam/blob/master/lib/commands/exam.js#L3), which I can't do much about, but at least this way it only gets required when `includedCommands` also runs.

cc @stefanpenner

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,13 +2,11 @@
 
 'use strict';
 
-var Commands = require('./lib/commands');
-
 module.exports = {
   name: 'ember-exam',
 
   includedCommands: function() {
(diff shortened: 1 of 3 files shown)


stefanpenner:
> Unfortunately, I think the bulk of slowness comes from requiring the Test Command, which I can't do much about, but at least this way it only gets required when includedCommands also runs.

awesome, i'll poke at test command itself and see if I can speed it up as well.

Issue #31, opened 2016-06-07
Reintroduce weighted splitting


trentmwillis:
This is a really nice feature to have so that we can help ensure the split groups are running in similar amounts of time. Unfortunately, however, it is not exactly as simple as it seems on the surface.

The `TestLoader` only knows about the name of the module it is loading, which means that it is unaware of any QUnit modules or tests that live within that module which makes it hard to associate a runtime number with the module.

One way to solve this issue is to load modules one-by-one. We can then use a `QUnit.testDone` callback to check when all tests have run and use that as the runtime for the module. Load the next module and repeat.

A couple thoughts on the above approach:
- Since the modules must be loaded one-by-one I'm unsure if this will break assumptions of any testing setups, so we may want it to be a separate flag to run in that mode
- Reporting the actual results and subsequently loading them in will require interfacing with Testem. The loading part needs to be investigated. It is likely that we will need to implement a custom Testem reporter (or a wrapper around the reporter).

mariokostelac:
Why is this feature totally different than splitting?

trentmwillis:
@mariokostelac this is different because it is possible to have unbalanced test suites with the current setup.

Consider a simplistic test suite that has 100 test files and all the odd numbered test files have 1 test case but all the even numbered test files have 10 test cases. If you do `--split=2` the first test group will have 50 tests total while the second test group will have 500 tests total.

These can also be imbalanced if you have some really slow tests in a specific module and really fast tests in other modules.

All that said, I have a new idea of how to solve this problem that I'll be attempting to implement this week/weekend.

mariokostelac:
That'd be great. I've tried `split` option with partitions and it was quite unbalanced. I did not check the number of tests in each case (because had some other issues with poorly written tests).
Is there any problem in loading all the tests and then just selecting which tests to write? Is that even possible with the way we write tests in ember?
Anyhow, if you need a hand, LMK. I am quite started in ember land, but seems so far I can learn a lot.

On `weighted` note. Would it be ever possible to extend the thing to give input file to override weights, per test? That would allow collection of test execution times, taking median and having even better balancing. Ofc, that collection and rebalancing weights would not be on `ember-exam`.

alexlafroscia:
Another thing to think about: I've notices some imbalances when splitting based on the fact that Acceptance tests take so much longer to run than unit tests. We split across 8 runners, and you can always tell which one had the most Acceptance tests because one of the 8 will take much longer than the rest.

[...]

PR #40, opened 2016-07-01
Temporarily undocument `--weighted`.

martndemus:
It had me confused if it worked or not. I propose to hide it until #31 lands.


trentmwillis:
Seems fair to me, should have probably done this before. Thanks!

Issue #48, opened 2016-07-26
Fix ember-cli-test-loader require


trentmwillis:
Pretty sure [this is going to break](https://github.com/trentmwillis/ember-exam/blob/43eb7603572241c28079ee9e17843dcdfe36a7d8/vendor/ember-exam/test-loader.js#L14) with the recent change of `ember-cli-test-loader` moving to NPM.

See a reference of the [fix in `ember-cli-qunit`](https://github.com/ember-cli/ember-cli-qunit/blob/7b45b1bb40468ef07c5f969013f7fb9b18d314d8/vendor/ember-cli-qunit/test-loader.js#L4-L10). Easy to do, just need someone to do it and verify behavior.

chriskrycho:
:raises hand: I'll see if I can take a swing at this sometime this weekend.

trentmwillis:
@chriskrycho thanks! Best way to verify this would probably be to upgrade the addon to use `ember-cli@2.7.0`

chriskrycho:
👍🏼 Taking a stab at it now!

cibernox:
@chriskrycho let me know if you need help, I wanted to update my app to 2.7 but I'm blocked by this.

chriskrycho:
@cibernox, if you want to pick up and run with the PR I started (#49), please do. I got started, but ran into a [test failure](https://travis-ci.org/trentmwillis/ember-exam/jobs/148574265#L458) that I didn't have time to chase down. That should hopefully have a lot of the package.json grunt work done, if nothing else.

trentmwillis:
Should be fixed now with `0.4.5`.

PR #49, opened 2016-07-30
WIP: Upgrade all deps to align with Ember 2.7.0.

chriskrycho:
Some stuff here is distinctly broken. Figured the easiest way to proceed asynchronously was to open a PR and iterate, though. Let me know if anything obvious jumps out. Otherwise, I'll come back to this as I have time over the next several days.

---

- [x] Implement fix from ember-cli-qunit for #48.
- [x] Upgrade all dependencies for Ember 2.7
- [ ] Fix now-failing tests

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,22 +13,8 @@ insert_final_newline = true
 indent_style = space
 indent_size = 2
 
-[*.js]
-indent_style = space
-indent_size = 2
-
 [*.hbs]
 insert_final_newline = false
-indent_style = space
-indent_size = 2
-
-[*.css]
-indent_style = space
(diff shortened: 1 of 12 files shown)


cspanring:
The test-loader path is also hardcoded in this `beforeEach` hook: https://github.com/trentmwillis/ember-exam/blob/master/tests/unit/test-loader-test.js#L5

However, fixing it there didn't solve the issue for me. I had more (global) test failures in the `Unit | test-loader` which I couldn't resolve (`TypeError: require is not a function` ??).

trentmwillis:
Merged via #51. Thanks for your work @chriskrycho!

PR #50, opened 2016-08-03
Fix in 2.7

cibernox:


cibernox:
@trentmwillis @chriskrycho I've tacking your PR and fixed the tests.

However I'm not sure if what I did is the proper fix. Should the tests (https://github.com/trentmwillis/ember-exam/pull/50/commits/984f77b1c2739a5742e7bac05ca4c90a0ee79f9e#diff-9a7b910d76a4ba883dc7d784e022c326R5) also load conditionally one package or the other?

cibernox:
I'm investigating the failure in node-land. I suspect that it's a core-object issue, but I think I'll need help for this.

trentmwillis:
I looked into this a bit. The Node failures are because we can no longer spy/stub super method calls.

I'm going to look into following a pattern similar to what Ember-CLI does for [the actual `test` command](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/commands/test-test.js).

trentmwillis:
Merged via #51. Thanks for your work @cibernox!

PR #63, opened 2016-10-28
Support specifying multiple partitions

### LINKED ISSUE #62: Parallel nested inside of partition?
I have a project that really needs refactoring, but currently has a very large number of tests that take a long time to run. I use CircleCI's parallel containers feature along with some custom code that I wrote that achieves the same effect as your partitioning. I suspect that I'm not maxing out the resources in each container, though, so what I'd love to try is partitioning my tests across containers, and then in each container parallelizing the tests in its partition.

Good idea? Silly idea? I'd probably be willing to take a crack at implementing it in a PR if you tell me if you think it's a good idea and help me figure out the API (CLI) design.

bendemboski:
The primary use case is splitting test modules across multiple CI containers, and having each of those CI containers run its list of tests in multiple parallel runs (you would specify split, parallel, and multiple partitions for each container). This can also be used for unbalanced partitioning, e.g. if you want one container to run 2/3 of the tests and the other run 1/3 (you would specify split and then different numbers of partitions for different containers).

Fixes #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,18 @@ ok 3 PhantomJS 1.9 - Exam Partition #2 - some other test
 
 You can also combine the `parallel` option with the `partition` option to split tests, and then recombine partitions into parallel runs. This would, for example, allow you to run tests in multiple CI containers and have each CI container parallelize its list of tests.
 
+For example, if you wanted to run your tests across two containers, but have one of them run twice as many tests as the other, and run them in parallel, you could do this:
+
+```bash
+# container 1
+ember exam --split=3 --partition=1 --partition=2 --parallel
+```
+
+```bash
+# container 2
+ember exam --split=3 --partition=3
+```
+
 Ember Exam will respect the `parallel` setting of your [Testem config file](https://github.com/testem/testem/blob/master/docs/config_file.md#config-level-options) while running tests in parallel. _Note that the default value for `parallel` in Testem is 1, which means you'll need a non-default value to actually see parallel behavior._
 
 _Note: You must be using Testem version `1.5.0` or greater for this feature to work properly._
--- a/README.md
+++ b/README.md
@@ -74,7 +74,11 @@ The `split` option allows you to specify a number of partitions greater than one
 $ ember exam --split=<num> --partition=<num>
 ```
 
-The `partition` option allows you to specify which test group to run after using the `split` option. It is one-indexed, so if you specify a split of 3, the last group you could run is 3 as well.
+The `partition` option allows you to specify which test group to run after using the `split` option. It is one-indexed, so if you specify a split of 3, the last group you could run is 3 as well. You can also run multiple partitions, e.g.:
+
+```bash
+$ ember exam --split=4 --partition=1 --partition=2
+```
 
 _Note: Ember Exam splits test by modifying the Ember-CLI `TestLoader`, which means that tests are split up according to AMD modules, so it is possible to have unbalanced partitions. For more info, see [issue #60](https://github.com/trentmwillis/ember-exam/issues/60)._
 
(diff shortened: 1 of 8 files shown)


trentmwillis:
Thanks for the work on this! A quick glance looks pretty good. I'll review in depth some time this weekend, if you could look into the codeclimate issue before then it'd be appreciated.

bendemboski:
I think this is all ready to go now.

README.md
```diff
@@ -98,6 +102,8 @@ ok 2 PhantomJS 1.9 - Exam Partition #3 - some other other test
 ok 3 PhantomJS 1.9 - Exam Partition #2 - some other test
 ```
 
+You can also combine the `parallel` option with the `partition` option to split tests, and then recombine partitions into parallel runs. This would, for example, allow you to run tests in multiple CI containers and have each CI container parallelize its list of tests.
+
```
trentmwillis: Can we add another code example here of how to use the partition option with parallel?

lib/commands/exam.js
```diff
@@ -63,9 +63,13 @@ module.exports = TestCommand.extend({
...
 
-      // Ignore the partition option when paralleling
+      // Ignore the partition option when paralleling (we'll fill it in later)
       if (!this.validator.shouldParallelize) {
-        commandOptions.query = addToQuery(commandOptions.query, '_partition', commandOptions.partition);
+        var partitions = commandOptions.partition;
+        if (partitions) {
+          partitions = partitions.join(',');
+        }
+        commandOptions.query = addToQuery(commandOptions.query, '_partitions', partitions);
```
trentmwillis: We should leave this as `_partition` for consistency with the CLI params. QUnit will automatically treat repeated URL params as an array. So if we do `tests?_partition=1&_partition=2` then you'll get `QUnit.urlParams._partition === ['1', '2']`.

lib/commands/exam.js
```diff
@@ -144,18 +156,20 @@ module.exports = TestCommand.extend({
 
   /**
    * Generates the test pages to use for parallel runs. For a given baseUrl,
-   * it appends the split count and a parition number as query params.
+   * it appends the total split count and the partition numbers each page is
+   * running as query params.
    *
    * @param {String} baseUrl
+   * @param {Array} partitions
```
trentmwillis: Should be `Array<Number>` right?

lib/utils/tests-options-validator.js
```diff
@@ -41,26 +41,57 @@ Object.defineProperty(TestsOptionsValidator.prototype, 'shouldSplit', {
...
+function validatePartitionsUnique(partitions) {
+  partitions = partitions.sort();
+  for (var i = 0; i < partitions.length - 1; i++) {
+    if (partitions[i] === partitions[i + 1]) {
+      throw new Error('You cannot specify the same partition twice.');
+    }
   }
 }
 
+
```
trentmwillis: Nit: delete extra line

lib/commands/exam.js
```diff
@@ -34,7 +34,7 @@ module.exports = TestCommand.extend({
 
   availableOptions: [
     { name: 'split',        type: Number,                  description: 'A number of files to split your tests across.' },
-    { name: 'partition',    type: Number,                  description: 'The number of the partition to run after splitting.' },
+    { name: 'partition',    type: [Array, String],         description: 'The number of the partition(s) to run after splitting.' },
```
trentmwillis: Shouldn't this be `type: [Array, Number]`?

vendor/ember-exam/test-loader.js
```diff
@@ -37,37 +37,54 @@ jQuery(document).ready(function() {
...
+      }
+
+      if (split < partition) {
+        throw new Error('You must specify partitions numbered less than or equal to your split value of ' + split);
+      } else  if (partition < 1) {
+        throw new Error('You must specify partitions numbered greater than 0');
+      }
+
+      var group = partition - 1;
+      tests = tests.concat(lintTestGroups[group]).concat(otherTestGroups[group]);
```
trentmwillis: `concat()` can acceptable multiple arrays at once, so let's do that instead.

vendor/ember-exam/test-loader.js
```diff
@@ -37,37 +37,54 @@ jQuery(document).ready(function() {
...
+    }
+    return tests;
```
trentmwillis: Nitpick: newline before this `return`.

bendemboski:
@trentmwillis feedback addressed

trentmwillis:
:tada: thanks for all your work on this!

bendemboski:
👍

Dhaulagiri:
@bendemboski @trentmwillis do you have an example of this working on circleci somewhere one could look at? I'm trying this in our circle setup but parallelizing tests within a container doesn't seem to make the overall test runs much faster.

(happy to move this conversation to community slack instead)

bendemboski:
@Dhaulagiri I don't have an example that I can make public. My guess is that the limiting factor in your test speed is CPU, rather than blocking on async events or something, so parallelizing within the same container won't speed anything up.

PR #67, opened 2016-11-18
Fixed issue using a single partition with a double digit number

patocallaghan:
If you have a command like `ember exam --split=10 --partition=10`, the command fails to run with a `You must specify partitions numbered greater than 0` error.

This is because `10` is been treated as a string when we get to https://github.com/trentmwillis/ember-exam/blob/master/vendor/ember-exam/test-loader.js#L72 and we try to loop over the number a digit at a time, i.e. first "1" and then "0"

### DIFF
--- a/tests/unit/test-loader-test.js
+++ b/tests/unit/test-loader-test.js
@@ -201,3 +201,29 @@ test('load works without non-lint tests', function(assert) {
     'test-4-test.jshint',
   ]);
 });
+
+test('load works with a double-digit single partition', function(assert) {
+  window.requirejs.entries = {
+    'test-1-test': true,
+    'test-2-test': true,
+    'test-3-test': true,
+    'test-4-test': true,
+    'test-5-test': true,
+    'test-6-test': true,
+    'test-7-test': true,
+    'test-8-test': true,
+    'test-9-test': true,
+    'test-10-test': true,
+  };
+
+  QUnit.urlParams = {
+    _partition: '10',
+    _split: 10,
+  };
+
+  this.TestLoader.load();
+
+  assert.deepEqual(this.requiredModules, [
+    'test-10-test',
+  ]);
+});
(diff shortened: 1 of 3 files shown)


tests/unit/test-loader-test.js
```diff
@@ -121,7 +121,7 @@ test('throws an error if partition isn\'t a number', function(assert) {
 
   assert.throws(() => {
     this.TestLoader.load();
-  }, /You must specify numbers for partition \(you specified 'foo'\)/);
+  }, /You must specify numbers for partition/);
```
patocallaghan: Rather than complicate things and try to keep a copy of the original `partition` before parsing I decided to remove the `(you specified ..)` part. Happy to add it back in if you think it should stay.

tests/unit/test-loader-test.js
```diff
@@ -201,3 +201,29 @@ test('load works without non-lint tests', function(assert) {
...
+    'test-5-test': true,
+    'test-6-test': true,
+    'test-7-test': true,
+    'test-8-test': true,
+    'test-9-test': true,
+    'test-10-test': true,
+  };
+
+  QUnit.urlParams = {
+    _partition: '10',
```
patocallaghan: Pass `_partition` as a string

tests/unit/test-loader-test.js
trentmwillis: I would prefer this as it is more useful for debugging.

patocallaghan:
Thanks @trentmwillis. Updated per your feedback!

trentmwillis:
Thanks! This looks good to me, but there are some CI failures. I think you just need to update the [test count for the Node tests](https://github.com/trentmwillis/ember-exam/blob/675a7f6a5e2d096cef08062fe34705034febba97/node-tests/acceptance/exam-test.js#L14).

patocallaghan:
Updated and now 💚..Thanks!

trentmwillis:
Great, thank you!

trentmwillis:
Released as `v0.5.3`

PR #68, opened 2016-11-18
Balance acceptance tests across partitions

bendemboski:
Acceptance tests generally take longer to run than unit/integration tests, so spread them evenly across partitions when splitting.

### DIFF
--- a/tests/unit/test-loader-test.js
+++ b/tests/unit/test-loader-test.js
@@ -18,6 +18,14 @@ module('Unit | test-loader', {
       'test-3-test.jshint': true,
       'test-4-test': true,
       'test-4-test.jshint': true,
+      'acceptance/test-1-test': true,
+      'acceptance/test-1-test.jshint': true,
+      'acceptance/test-2-test': true,
+      'acceptance/test-2-test.jshint': true,
+      'myapp/acceptance/test-3-test': true,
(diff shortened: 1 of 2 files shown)


bendemboski:
@trentmwillis what are your thoughts? Just an idea that I had that seemed almost as fast to code up and PR than file an issue for discussion. So let me know if you think this makes sense or not.

trentmwillis:
Thanks for the PR!

Generally, if you're not doing custom build stuff, then your tests will load in alphabetical order. Additionally, the partitioning algorithm is very naive and puts every `n`th module into partition `n`. Therefore, you should actually wind up with a fairly balanced split of acceptance, integration, and unit tests by default.

That said, an earlier iteration of this addon attempted to "weight" each type of test by doing something similar to what you have proposed here; acceptance tests weighed more than integration tests weighed more than unit tests. This proves to be very fragile and just as easily becomes unbalanced, because each module can have varying numbers of tests within them. Thus, an acceptance test module with 1 test is much less costly than a module with 10 or even 100 tests.

I do want to support "weighted" splitting at some point in the future, but it needs to be a robust solution that is actually based on the historical runtime of each module. I discuss and am tracking this feature in https://github.com/trentmwillis/ember-exam/issues/31.

Closing this for now; would love help with #31 though if you're interested! I'm not 100% sure how to solve it at this point in time.

PR #87, opened 2017-05-09
Add Node LTS versions to .travis.yml

cbrock:
@trentmwillis Seems like it would be good for Travis to be running against the current Node LTS versions ([currently 4.x and 6.x](https://github.com/nodejs/LTS)).

Additionally, since Node 0.12 was EOL'ed last year, I removed that.

For what it's worth, I see test failures locally when running master against Node 4.7.2 so I'm curious if those failures will surface in Travis as well.

Happy to adjust the PR if you have any feedback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 ---
 language: node_js
 node_js:
-  - "0.12"
   - "4"
   - "6"
   - "stable"
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,8 @@ cache:
     - node_modules
 
 env:
-  - TEST_FRAMEWORK=ember-cli-qunit EMBER_TRY_SCENARIO=default
-  - TEST_FRAMEWORK=ember-cli-qunit EMBER_TRY_SCENARIO=ember-release
-  - TEST_FRAMEWORK=ember-cli-qunit EMBER_TRY_SCENARIO=ember-beta
-  - TEST_FRAMEWORK=ember-cli-qunit EMBER_TRY_SCENARIO=ember-canary
-  - TEST_FRAMEWORK=ember-cli-mocha EMBER_TRY_SCENARIO=default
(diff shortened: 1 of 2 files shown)


cbrock:
@trentmwillis The failures seem to be related to the "partition" number no longer existing in the test output, e.g.

```
1) Acceptance | Exam Command Split splits the test suite but only runs the first partition:

      Uncaught AssertionError: output has partition #1
      + expected - actual

      -false
      +true
```

Any idea what dependency change could have caused this? I'd be happy to update my PR to include a fix, but would appreciate a nudge in the right direction.

trentmwillis:
@cbrock interesting, I'm not aware of anything that has changed that should cause those failures. I've been using Ember Exam with Node 6 and new Ember apps relatively frequently. I'll try to dig in soon. Also, thanks for the PR! This definitely should be done.

.travis.yml
```diff
@@ -1,7 +1,8 @@
 ---
 language: node_js
 node_js:
-  - "0.12"
+  - "4"
+  - "6"
```
trentmwillis: Can we also include `"stable"` here?

I'd also like to update the actual test command that gets runs from an individual ember-try scenario to use `ember try:each` instead. That way we don't eat up tons of Travis processes.

trentmwillis:
I believe the test failures here are due to the same issue as in https://github.com/trentmwillis/ember-exam/issues/88. I'm going to lock the Testem version here until the fix is released.

.travis.yml
cbrock: @trentmwillis Good suggestion - I've added `"stable"`.

I think I've understood your `try:each` suggestion correctly, but if you wouldn't mind verifying the last commit in this PR I'd appreciate it!

cbrock:
@trentmwillis Looks like the Testem fix is in and the test suite now passes for me locally and in Travis.

PR #107, opened 2017-09-10
Improve installation instructions

mariokostelac:
I think I would not open https://github.com/trentmwillis/ember-exam/issues/106 if instructions were looking like this :).


mariokostelac:
Thanks @trentmwillis :)

PR #114, opened 2017-10-05
Remove `#` from test output. Fixes #112

### LINKED ISSUE #112: Ember Exam breaks TAP reporting
For users using a CI that parses testem output as [TAP 13](https://testanything.org/tap-version-13-specification.html) reporting, the addition of `Exam Partition #1 -` to the output breaks the parsing, effectively telling the TAP consumer that there is only one test named "Exam Partition" (the `#` signals that a directive is coming).

**Possible Solutions:**

1. Remove the `#` (not tested, but I think that's all that is required to fix it)
2. Move the`Exam Partition` stuff to after the test name
3. Allow users to configure what Ember Exam adds to the output
3. Allow users to turn off appending anything to the output

I would be happy to do the work, I just need some guidance on the preferred solution.

luketheobscure:


luketheobscure:
Fixed the tests, and I went ahead and updated the readme to remove any references to to the `#`.

PR #119, opened 2017-10-23
first take on per-test qunit splitting

mariokostelac:
This is an example of how we could support per-test test splitting for
qunit. I would like to get feedback on the approach and whether it has
some future so I could continue working on it.
It breaks mocha so I am interested in thoughts on that as well.

Tests failing: unit tests of TestLoader.


mariokostelac:
Just to note, I am willing to work on getting it done for Mocha as well if this approach is ok with you.

mariokostelac:
No problem, thanks for taking time to comment :). Looking forward to make it work for everybody.

mariokostelac:
I will try to get them passing next week. It is possible I will need a little bit of help. If that happens, will communicate through comments here.

PR #121, opened 2017-11-22
set process.env.EMBER_EXAM_SPLIT

### LINKED ISSUE #120: set env var with value of split
It would be awesome if using `--split=5` would set `process.env.EMBER_EXAM_SPLIT = 5`. Then testem.js setup would be as clean as

```js
  parallel: process.env.EMBER_EXAM_SPLIT,
```

Thoughts?

kellyselden:
Closes #120

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,9 @@ ember exam --split=3 --partition=3
 
 **Note 1**: _Ember Exam will respect the `parallel` setting of your [Testem config file](https://github.com/testem/testem/blob/master/docs/config_file.md#config-level-options) while running tests in parallel. The default value for `parallel` in Testem is 1, which means you'll need a non-default value to actually see parallel behavior._
 
-**Note 2**: _You must be using Testem version `1.5.0` or greater for this feature to work properly._
+**Note 2**: _Ember Exam sets `process.env.EMBER_EXAM_SPLIT_COUNT` for convenience. You can use this in your Testem file._
+
+**Note 3**: _You must be using Testem version `1.5.0` or greater for this feature to work properly._
 
 ## Advanced Configuration
 
--- a/node-tests/unit/commands/exam-test.js
+++ b/node-tests/unit/commands/exam-test.js
@@ -110,6 +110,12 @@ describe('ExamCommand', function() {
         assert.equal(called.testRunOptions.query, '_split=2&_partition=2&_weighted');
       });
     });
+
+    it('should set split env var', function() {
+      return command.run({ split: 5 }).then(function() {
+        assert.equal(process.env.EMBER_EXAM_SPLIT_COUNT, '5');
+      });
+    });
   });
 
   describe('_generateCustomConfigs', function() {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -157,7 +157,9 @@ ember exam --split=3 --partition=3
 
 **Note 1**: _Ember Exam will respect the `parallel` setting of your [Testem config file](https://github.com/testem/testem/blob/master/docs/config_file.md#config-level-options) while running tests in parallel. The default value for `parallel` in Testem is 1, which means you'll need a non-default value to actually see parallel behavior._
 
-**Note 2**: _You must be using Testem version `1.5.0` or greater for this feature to work properly._
+**Note 2**: _Ember Exam sets `process.env.EMBER_EXAM_SPLIT` for convenience. You can use this in your Testem file._
```
rwjblue: I'd slightly prefer `EMBER_EXAM_SPLIT_COUNT`.  I find `EMBER_EXAM_SPLIT` _slightly_ confusing (is it the total number of splits, or the current partition number?)...

kellyselden: @trentmwillis Thoughts on the matter?

kellyselden: I think I prefer `EMBER_EXAM_SPLIT` because it maps well to `--split`, as well as a future `EMBER_EXAM_PARTITION` mapping to `--partition`, but I'm cool with whatever the consensus is.

trentmwillis: @kellyselden I agree with you. I can understand @rwjblue's point, but if we mirror other options in the future, I'd like a consistent naming pattern.

node-tests/unit/commands/exam-test.js
```diff
@@ -110,6 +110,12 @@ describe('ExamCommand', function() {
         assert.equal(called.testRunOptions.query, '_split=2&_partition=2&_weighted');
       });
     });
+
+    it('should set split env var', function() {
+      return command.run({ split: 5 }).then(function() {
+        assert.equal(process.env.EMBER_EXAM_SPLIT, '5');
```
trentmwillis: CodeClimate doesn't like the similarity between this test and one of the other ones above. Could you look into combining them? Since they both only test the `split` option, it makes sense to merge them and just have multiple assertions.

stefanpenner:
I've rebased this PR. @rwjblue you had a concern above, are you ok with moving forward with `EMBER_EXAM_SPLIT` or should we strongly consider `EMBER_EXAM_SPLIT_COUNT`

cc @kellyselden

rwjblue:
I do still prefer split count, but not massively...

stefanpenner:
@kellyselden can you get behind that? As it is a env var, specifically for power users, I can get behind the extra clarity/verbosity.

kellyselden:
Works for me.

kellyselden:
renamed to EMBER_EXAM_SPLIT_COUNT

stefanpenner:
released as v2.0.1 🎉

PR #128, opened 2018-02-02
Update to 2.18 family

cibernox:


cibernox:
I'm not sure why codeclimate is failing. It is supposed to run eslint but with with the new config in .eslint.js there seems to not have any warning. Any idea?

stefanpenner:
Sorry for letting this linger, i have a PR getting us to latest here which will supercede this: https://github.com/ember-cli/ember-exam/pull/161

PR #130, opened 2018-06-11
chore: migrate to cli-table3

DanielRuf:


package.json
```diff
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "chalk": "^2.1.0",
-    "cli-table2": "^0.2.0",
+    "cli-table3": "^0.5.0",
```
rwjblue: I think you will need to use 0.4.0 or update `.travis.yml`/`engines`/etc to drop Node 4.

DanielRuf: I think dropping Node.js 4 is better as it reached its EOL.

rwjblue: Totally agree, but it generally requires a major version bump so I figured I'd point out the option...

Turbo87: I tend to agree. Let's us v0.4.0 for now, and then we can drop Node 4 in a separate step :)

DanielRuf: Changed it to 0.4.0.

stefanpenner:
we are about to do a major version bump, so I will rebase this now :)

PR #147, opened 2018-11-07
get rid of ember-cli-qunit dependency

choheekim:
This pr contains three changes:
1. Removing dependency in ember-cli-qunit - ember-cli-qunit is getting deprecated and Ember Exam's using ember-cli-qunit to invoke ember-qunit as a wrapper.
2. Disable sandbox when running Chrome inside a container in testem.js
3. Pass { setupEmberOnerrorValidation: false } param to disable running Ember.onerror validation test in ember-exam

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
     - $HOME/.npm
 
 env:
-  - TEST_FRAMEWORK=ember-cli-qunit
+  - TEST_FRAMEWORK=ember-qunit
   - TEST_FRAMEWORK=ember-cli-mocha
 
 matrix:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,9 +1,9 @@
 ---
 language: node_js
 node_js:
-  - "4"
```
rwjblue: This seems somewhat unrelated, and generally considered breaking (since we would no longer be testing Node 4).

Could you make this change in a separate PR and update the `package.json`'s `engines` field? That will help keep things together for the changelog...

choheekim: Open a pr with the change:
https://github.com/ember-cli/ember-exam/pull/148

rwjblue:
OK, should be ready with a rebase...

PR #136, opened 2018-10-03
[Test Load Balancing] Validate `load-balance` option and enable to load test balanced

choheekim:
This pr contains two changes:

1) The weight-test-modules allows ordering a list of AMD modules by slowest to fastest test type where we assume that acceptance tests are the slowest test type, with integration, unit tests and lint test being fast and faster and fastest. Sorting the list of AMD modules will be functioned by default when load-balance option is specified to provide better utilization of browser parallelization.

2) With load-balance options enabled that the testLoader requires and unsees an AMD module as requested instead of testLoader loads a list of AMD modules all once. As the order of AMD modules ran per browser is not consistent with the load-balance option modules ran per browser are tracked and when all tests are complete it creates a json file which maps from a browser id to a list of AMD modules ran on the browser.

Issue #134, opened 2018-08-28
RFC: Re-introduce weighted test Ordering with Test Load Balancing


step2yeung:
This RFC outlines a plan to re-introduce weighted test ordering by test types with test load balancing in Ember Exam for Ember applications. The goals of this RFC is to:
- Enable scalable test executions 
- Optimize the speed of test runs no matter the number of tests
- Decrease test start time by loading tests on demand

[Rendered](https://gist.github.com/step2yeung/94c01cc4dcfae5af2c2d64464233a7da)

trentmwillis:
Thanks for this very detailed proposal @step2yeung! Overall, I am very :+1: on this idea. Great enumeration of the constraints and problem space for this feature.

That said, after reading I'm not sure this proposed solution actually fixes the stated problem.

In particular, since `--load-balance` applies to a single partition (example from the RFC):

```
--split=2 --partition=1 --load-balance=20 --weighted
--split=2 --partition=2 --load-balance=20 --weighted
```

This means that the actual partitioning will still be done statically and thus the partitions will wind up being just as unbalanced as they are today.

What I would expect instead is to have a command like the original `--weighted` flag, maybe:

```
--split=2 --parallel --balanced
```

This would then allow you to balance all the modules into equally weighted partitions as proposed in the RFC.

----

That said, I can understand the desire to have different, statically defined partitions that you can run on separate CI boxes, but that is solving a different problem from what is stated here.

step2yeung:
Thanks for the feedback @trentmwillis !

You bring up a good point, when `--load-balance` is used with `--split`, it really just balances the tests within partition 1 against 20 number of browsers. 

In the example you refer to, the concept of `partition` applies to a CI box instead of a browser. So it could be the case that partition 1 on CI box 1 finishes before partition 2 on CI box2 (or the other way around). So, it is not truly balanced when looking at all the runs as a whole.

The above is true when running on separate CI boxes. But when it comes to a single CI box, I think it is still valid:

If I am understanding correctly `--split=2 --parallel --balanced` would sort the modules by weight with `--balanced`, then `--split=2 --parallel` would split the list of modules into two partitions, which would be ran on two browsers.
The proposed solution's equivalent would be `--weighted --load-balance=2`

More thought is needed to truly balance partitions against multiple CI boxes, but I wanted to preserve this functionality of running different partitions on separate CI box for the `--load-balance` feature. Do you feel this is a good step forward?

trentmwillis:
[...]

PR #159, opened 2018-11-15
Update README.md with instruction for usage of test load balance

choheekim:
This pr is to update README.md to provide an instruction of new features in this pr - https://github.com/ember-cli/ember-exam/pull/136.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,13 +182,13 @@ ember exam --split=3 --partition=3 --parallel
 $ ember exam --parallel=<num> --load-balance
 ```
 
-`load-balance` allows you to load balance test files against. It will order the test files by test type and load balance the ordered test files between the browsers dynamically rather than statically. `parallel` must be used along with `load-balance` to specify a number of browser(s) and `load-balance` can not be used without `parallel` being specified a number of browser(s).
+The `load-balance` option allows you to load balance test files against multiple browsers. It will order the test files by test types, e.g. acceptance | integration | unit | eslint, and load balance the ordered test files between the browsers dynamically rather than statically.
+**Note:** parallel must be used along with load-balance to specify a number of browser(s)
 
-This option was added to version 1.1 to address execution performance when running against a large test suite.
+The `load-balance` option was added to version 1.1 to address execution performance when running against a large test suite.
 
-A way of communication between server and browser to send and receive a test module to load is promise based. Its timeout value is set to be 2 seconds and the timeout can be customized by user by adding `asyncTimeout=[timeout]` in passing url or adding to a testem config.
-
-For example, if you specify load-balance and parallel equals 3, then three browser instances will be created and the output will look something like:
+Web browsers and servers communicate via promise in order to send and receive a test file. The promise timeout value is set to be 2 seconds, and the timeout can be customized by adding asyncTimeout=[timeout] as a querystring param in the test URL or adding to a testem config.
+For example, if you specify `load-balance` and `parallel` equals 3, then three browser instances will be created and the output will look something like:
 
 ```bash
 # ember exam --parallel=3 --load-balance
@@ -203,7 +203,7 @@ You can also specify the `split` and `partition` options with `load-balance` to 
 $ ember exam --split=<num> --partition=<num> --parallel=<num> --load-balance
 ```
 
-This command will split test files and load-balance tests from the specified partition across the browsers. For example --split=2 -partition=1 --parallel=3 --load-balance, the complete list of test files are split into two halves. With the first half of the list load balanced against three browsers. The output will look something like below:
+This command will split test files and load-balance tests from the specified partition across the browsers. For example `ember exam --split=2 -partition=1 --parallel=3 --load-balance`, the complete list of test files are split into two halves. With the first half of the list load balanced against three browsers. The output will look something like below:
 
 ```bash
 # ember exam --split=2 --partition=1 --parallel=3 --load-balance
@@ -213,66 +213,68 @@ ok 3 Chrome 66.0 - Exam Partition 1 - browser Id 3 - some the other test
 ```
 
 
-`load-balance` is currently supported in CI mode and due to the reason `no-launch` can not be passed with `load-balance`.
+**Important information on load-balance**
 
-**Note 1** _You must be using ember-cli version 3.2.0 or greater for load-balnce and test failure reproduction features to work properly._
-**Note 2** _You must be using ember-qunit version 4.1.1 or greater for this feature to work properly._
-**Note 3** _You must be using qunit version 2.8.0 or greater for this feature to work properly._
-**Note 4** _This feature is not currently supported by Mocha._
+1. The load-balance option is currently only supported in CI mode and due to the reason no-launch can not be used with load-balance.
+2. You must be using ember-cli version 3.2.0 or greater for load-balnce and test failure reproduction features to work properly.
+3. You must be using ember-qunit version 4.1.1 or greater for this feature to work properly.
+4. You must be using qunit version 2.8.0 or greater for this feature to work properly.
+5. This feature is not currently supported by Mocha.
 
 ##### Test Failure Reproduction
 
-Due to the dynamic nature of the `load-balance` option, test file execution order could vary between runs. To reproduce a past test execution the execution must be recorded and passing `--write-execution-file`, in short `--wef` allows generating a JSON file which enables to rerun the past test execution. The option is only allowed when `load-balance` is passed.
+Due to the dynamic nature of the load-balance option, test file execution order can vary between runs. In order to reproduce a past test execution, the execution must be recorded via passing --write-execution-file or --wef, which allows generating a JSON file that enables rerunning the past test execution. The option is only allowed when load-balance is passed.
 
 ```bash
-# The command will load test balanced with serve <num> of browser(s). After the test suite execution It will generate a test-execution json file.
+# The command will load in test balanced mode with <num> of browser(s). After the test suite execution, it will generate a test-execution json file.
 $ ember exam --parallel=<num> --load-balance --wef
 $ ember exam --parallel=<num> --load-balance --write-execution-file
 ```
 
-The file is stored in the root directory and the naming structure is `test-execution-YYYY-MM-DD--HH-MM-SS.json`.
-To replay the test execution for one particular browser, do the following:
+The file is stored in the root directory and the naming structure is `test-execution-<timestamp>.json`.
+To replay the test execution for particular browser(s), do the following:
 
 ```bash
-# The command will read test execution file passed for `replay-execution` and execute a browser Id(s) specified for `rplay-browser`
+# The command will read a test execution file specified for `replay-execution` and execute a browser Id(s) from `replay-browser`
 $ ember exam --replay-execution=[string] --replay-browser=[num]
 ```
 
-`replay-execution` allows you to specify a path to the json file to run execution against and `replay-browser` is to specify browser IDs to execute.
+`replay-execution` allows you to specify a path to the json file to run execution against and `replay-browser` is to specify browser ID(s) to execute.
 
 ```bash
-# The command will read test-execution-000000.json and load the list of module mapped to browserId 1
+# The command will read test-execution-000000.json and load the list of modules mapped to browserId 1
 $ ember exam --replay-execution=test-execution-000000.json --replay-browser=1
 ```
 
 The above command will read `test-execution-000000.json` and load the list of modules which is mapped by browser ID #1.
 
-`replay-browser` can be an array of broweser IDs. For instance `--replay-broweser=1,2` will bring two browsers and execute a list of modules which ran by previous browser #1 and #2.
+`replay-browser` can be an array of browser IDs. For instance `--replay-broweser=1,2` will start two browsers and execute a list of modules which were previously run by browsers #1 and #2.
 
 ```bash
 # The command will read test-execution-000000.json and load the list of module mapped to browserId 1 and 2
 $ ember exam --replay-execution=test-execution-000000.json --replay-browser=1,2
 ```
 
-When `replay-browser` value is not specified its value is set to be a value of `failedBrowser` from the specified `replay-execution` file.
+When `replay-browser` value is not specified it will execute browserId(s) read from `failedBrowser` in the test execution file.
 
 ```bash
-# The command will read test-execution-000000.json and load the list of module mapped to browserIds from failedBrowser in the json file.
+# The command will read test-execution-000000.json and load the list of modules mapped to browserIds from failedBrowser in the json file.
 $ ember exam --replay-execution=test-execution-000000.json
 ```
 
-When `replay-browser` value is not specified and there is no value for `failedBrowser` in the json file it will execute all list of modules.
+When `replay-browser` value is not specified and there is no value for `failedBrowser` in the json file it will rerun all list of modules.
 
 ```bash
 # The command will read test-execution-000000.json and load the list of module mapped to all browserIds when failedBrowser is none in the json file
 $ ember exam --replay-execution=test-execution-000000.json
 ```
 
-**Note 1** _You must be using ember-cli version 3.2.0 or greater for load-balnce and test failure reproduction features to work properly._
-**Note 2** _You must be using ember-qunit version 4.1.1 or greater for this feature to work properly._
-**Note 3** _You must be using qunit version 2.8.0 or greater for this feature to work properly._
-**Note 4** _This feature is not currently supported by Mocha._
+**Important information on replay-execution and replay-browser**
 
+1. You must be using ember-cli version 3.2.0 or greater for load-balnce and test failure reproduction features to work properly.
+2. You must be using ember-qunit version 4.1.1 or greater for this feature to work properly.
+3. You must be using qunit version 2.8.0 or greater for this feature to work properly.
+4. This feature is not currently supported by Mocha.
 
 ## Advanced Configuration
 
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Code Climate](https://codeclimate.com/github/trentmwillis/ember-exam/badges/gpa.svg)](https://codeclimate.com/github/trentmwillis/ember-exam)
 [![Node Test Coverage](https://codeclimate.com/github/trentmwillis/ember-exam/badges/coverage.svg)](https://codeclimate.com/github/trentmwillis/ember-exam/coverage)
 
-Ember Exam is an addon to allow you more control over how you run your tests when used in conjunction with [Ember QUnit](https://github.com/emberjs/ember-qunit) or [Ember CLI Mocha](https://github.com/ember-cli/ember-cli-mocha). It provides the ability to randomize, split, and parallelize your test suite by adding a more robust CLI command.
+Ember Exam is an addon to allow you more control over how you run your tests when used in conjunction with [Ember QUnit](https://github.com/emberjs/ember-qunit) or [Ember CLI Mocha](https://github.com/ember-cli/ember-cli-mocha). It provides the ability to randomize, split, parallelize, and load-balance your test suite by adding a more robust CLI command.
 
 It started as a way to help reduce flaky tests and encourage healthy test driven development. It's like [Head & Shoulders](http://www.headandshoulders.com/) for your tests!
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,6 +165,67 @@ ember exam --split=3 --partition=3
...
+
+```bash
+# don’t do this
+# The number of browser instances = --server --no-launch
+$ ember exam --load-balance --server --no-launch
+
+# do this
+$ ember exam --server --no-launch --load-balance
+```
+
```
step2yeung: Can we add a note here to point out the dynamic nature of `load-balance` will likely mean test file execution order could vary? So users of the command won't be surprised, and probably helps the flow into the next section.

eg:
Due to the dynamic nature of the `load-balance` option, test file execution order could vary between runs.

README.md
```diff
@@ -159,6 +165,67 @@ ember exam --split=3 --partition=3
...
+##### Test Failure Reproduction
+
+Since Ember Exam with `load-balance` loads modules one at a time the order of executing modules is not deterministic. For this reason a test execution with `load-balance` creates a json file to track which browser executes which modules in which order. To replay the tests for one particular browser, do the following:
```
step2yeung: Since we mentioned the json file here, can we also add a sentence noting where the file gets created and the naming structure?

Also, a small nitpick: ` To replay the test execution for one particular browser`

README.md
```diff
@@ -159,6 +165,67 @@ ember exam --split=3 --partition=3
...
+
+```bash
+$ ember exam --replay-execution=[string] --replay-browser=[num]
+```
+
+`replay-execution` allows you to specify a json file to run execution against and `replay-browser` is to specify browser IDs to run from the json file.
```
step2yeung: Suggestion: `replay-execution` allows you to specify a path to the json file to run execution against and `replay-browser` is to specify browser IDs to execute.

README.md
```diff
@@ -159,6 +165,67 @@ ember exam --split=3 --partition=3
...
+
+```bash
+$ ember exam --replay-execution=test-execution-000000.json --replay-browser=1
+```
+
+The above command will read `test-execution-000000.json` and load the list of modules which is mapped by browser ID #1.
+
+
+**Note 1** _You must be using ember-cli version 3.2.0 or greater for this feature to work properly._
```
step2yeung: Are these notes(all 4) for `load-balance` in general, or only for `replay execution` ?

choheekim: it should be for both `load-balance` and `replay-execution`

README.md
```diff
@@ -159,6 +180,95 @@ ember exam --split=3 --partition=3
...
+To replay the test execution for one particular browser, do the following:
+
+```bash
+$ ember exam --replay-execution=[string] --replay-browser=[num]
+```
+
+`replay-execution` allows you to specify a path to the json file to run execution against and `replay-browser` is to specify browser IDs to execute.
+
+```bash
+$ ember exam --replay-execution=test-execution-000000.json --replay-browser=1
```
stefanpenner: can one provide `--replay-browser=1,2,3` to specify more then 1 browser (but not all) to replay?

If not, that is fine, it could easily be described as out-of-scope of an initial implementation.

README.md
```diff
@@ -159,6 +180,95 @@ ember exam --split=3 --partition=3
 
 **Note 2**: _You must be using Testem version `1.5.0` or greater for this feature to work properly._
 
+### Test Load Balancing
+
+```bash
+$ ember exam --load-balance=<num>
```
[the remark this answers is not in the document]
step2yeung: @stefanpenner , let me help answer this one:

Your suggestion was initially how the arguments were designed 😄. From the #134 feedback we went with the proposed change that we believe simplifies the commands.

For context, lets look at how ember-exam can be used today:
1) `ember exam` —> same as `ember test`
2) `ember exam --split=6` —> spawns 1 browser, splitting the test suite into 6 partitions, run partition 1 only
3) `ember exam --split=6 --partition=1,2,3` —> spawns 1 browser, splitting the test suite into 6 partitions, run partition 1,2,3
4) `ember exam --split=6 --parallel` —> spawns 6 browsers, splitting the test suite into 6 partitions, run all 6 partitions
5) `ember exam --split=6 --partition=1,2,3 --parallel` —> spawns 3 browsers, splitting the test suite into 6 partitions, run partitions 1, 2, 3 in machine 1
6) `ember exam --split=6 --partition=4,5,6 --parallel` —> spawns 3 browsers, splitting the test suite into 6 partitions, run partitions 4, 5, 6 in machine 2

The number of browsers spawned is controlled by different parameters depending on the combination of commands used, which we thought was a bit convoluted.
eg. 2) and 3) without using —parallel, only 1 browser is spawned.
     4) if parallel is used without partitions, split determines the number of browsers.
     5) and 6) if partitions are passed, the number of partitions determines the number of browsers.


Proposed ember exam with test load balancing:
1’) `ember exam` —> basically run ember test
2’) `ember exam --load-balance=6` —> bring up 6 browser, sharing the full test suite with 6 browsers
3’) `ember exam --load-balance=6 --split=2 --partition=1` —> bring 6 browsers, splitting the test suite into 2 partitions, run partitions 1 in machine 1
4’) `ember exam --load-balance=6 --split=2 --partition=2` —> bring 6 browsers, splitting the test suite into 2 partitions, run partitions 2 in machine 1

Using `--load-balance=<number>` allows a single argument that set the number of browser to spawn.
This removes the need for `parallel`, because the load-balance value always define the number of browsers to run in parallel.
The use case of 2) to spawns a single browser to run one partition effectively gets replaced with `replay-execution`.
And 3) that spawns a single browser to run multiple partitions is not supported by `load-balance`.

Another important change is the use of `split` and `partition`, which predefines partitions of tests used in two ways in today’s ember-exam:
Firstly, it allows browsers to execute a unique subset of tests.
Secondly, this unique subset of tests can be defined to run across multiple CI machines.

With `load-balance`, predefined partitions of tests is replaced with dynamically load balanced of tests, this removes the need for `split` and `partitions` to split up the test suite in a static way *for each browser.*
The ability to run tests across CI machines is still preserved, by using `split` and `partition` to predefine partitions of tests for *a CI machine*.
Think of `split` as number of machines, and `partition` as identifier of the machine, as demonstrated by 3’) and 4’)

The run test across CI machine command gets a bit confusing if `--load-balance` is only a boolean:
A) `ember exam --load-balance=true --split=6 --partition=1,2,3 --parallel` —> bring 3 browsers, splitting the test suite into 6 partitions, load balance the tests from partitions 1, 2, 3 on machine 1
B) `ember exam --load-balance=true --split=6 --partition=4,5,6 --parallel` —> bring 3 browsers, splitting the test suite into 6 partitions, load balance the tests from partitions 4, 5, 6 on machine 2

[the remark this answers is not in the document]
stefanpenner: Thanks for the detailed explanation. The way I can reconcile these constraints, this is as follows:

*note:  just because what described bellow makes sense to me, doesn't mean it makes sense to a broader audience, I am sharing it to see if it resonates and as a basis for further discussion.*

* `--parallel` starts taking an optional argument, which represents the amount fo parallelism. This number can be `> 0` although higher numbers may be underutilized, with the wrong scheduling type.
* `--parallel` defaults to `0`, which simply mimics todays functionality
* `--load-balance` or `--share-work` or `--scheduler='stealing'` becomes a way to choose the mode of scheduling. This would ensure the work-stealing scheduler is opt-in.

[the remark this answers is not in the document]
step2yeung: Thanks for the feedback @stefanpenner 

After some discussions offline, I feel your suggestions for the commands make sense going forward given:
- the possibility of making ember test adopt ember-exam's functionality
- load balancing become the default behavior, so the `--load-balance` command can go away

Taking your suggestion, the proposed commands will behave the same way as it does currently without `load-balance`. But with `load-balance`, it will introduce the new way to use `parallel` option to define the amount of parallelism.

Commands | Descriptions
-- | --
ember exam --split=6 | 1 browser, splitting the test suite into 6 partitions, run partition 1 only
ember exam --split=6 --partition=1,2,3 | 1 browser, splitting the test suite into 6 partitions, run partition 1,2,3
ember exam --split=6 --parallel | 6 browsers, splitting the test suite into 6 partitions, run all 6 partitions
ember exam --split=6 --partition=1,2,3 --parallel | 3 browsers, splitting the test suite into 6 partitions, run partitions 1, 2, 3 in machine 1
ember exam --split=6 --partition=4,5,6 --parallel | 3 browsers, splitting the test suite into 6 partitions, run partitions 4, 5, 6 in machine 2
  |  
ember exam --load-balance --parallel=6 | 6 browsers, sharing the full test suite with 6 browsers
ember exam --load-balance --split=2 --partition=1 --parallel=6 | 6 browsers, splitting the test suite into 2 partitions, run partitions 1 in machine 1
ember exam --load-balance --split=2 --partition=2 --parallel=6 | 6 browsers, splitting the test suite into 2 partitions, run partitions 2 in machine 1
  |  
ember exam --load-balance | 1 browsers, run full test suite using load-balance method
ember exam --load-balance --split=2 --partition=1 | 1 browser, splitting the test suite into 2 partitions, run partitions 1 in machine 1
ember exam --load-balance --split=2 --partition=2 | 1 browser, splitting the test suite into 2 partitions, run partitions 2 in machine 2


When `load-balance` becomes the default, the commands can be greatly simplified: (with some renaming of options like `split` to `partition` and `partition` to `run-part`)

Command | Descriptions
-- | --
ember exam | same as ember test
ember exam --filter='acceptance' | test with filter
ember exam --server | run test in server/dev mode
ember exam --random | run tests in random order
ember exam:iterate <num> |  
  |  
  | ** test distribution defaulted to load-balance
ember exam --parallel=6 | 6 browsers, sharing the full test suite with 6 browsers
ember exam --parallel=6 --partition=2 --run-part=1 | 6 browsers, splitting the test suite into 2 partitions, run part 1 in machine 1
ember exam --parallel=6 --partition=2 --run-part=2 | 6 browsers, splitting the test suite into 2 partitions, run part 2 in machine 1

README.md
```diff
@@ -107,7 +128,7 @@ _Note: This feature is not currently supported by Mocha._
 $ ember exam --split=<num>
 ```
 
-The `split` option allows you to specify a number of partitions greater than one to spread your tests across. Ember Exam will then proceed to run the first batch of tests.
+The `split` option allows you to specify the number of partitions greater than one to spread your tests across. Ember Exam will then proceed to run the first batch of tests.
```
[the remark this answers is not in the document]
scalvert: Seems grammatically more correct?

step2yeung:
Had more discuss about the UX/CLI with @choheekim and @scalvert and wanted to share what we discussed:

There are two problems with the current CLI of ember exam:
 - the ‘parallel’ command is convoluted 
- the naming of some of the commands are not straight forward

And with the load-balance changes, some commands can be simplified in both meaning and usage. 

Existing commands:
**‘Split’** splits the full list of test modules into a number of partitions set by the input value, but ‘split’ feels lacking in meaning when looking at the command by itself.
**‘Parallel’** is a boolean to run tests on multiple browsers. The number of browsers to invoke depending on the combination of commands: ‘split’, ‘partition’
**‘Partition’** accepts a numeric value, specifying the partition to execute.

 Sample usage: 
ember exam --split=10   // invoke 1 browser, running partition 1
ember exam --split=10 --parallel  // invoke 10 browsers defined by split, each running its own partition
ember exam --split=10 --partition=1 --partition=2 --partition3  // invoke 1 browsers, running tests defined in partition 1,2 and 3 
ember exam --split=10 --partition=1 --partition=2 --partition3 --parallel  // invoke 3 browsers, defined by the number of partitions specified in the command, each browser runs a unique partition


Newly proposed commands:
 **‘Partition-count’** replacing ‘split’ 
**‘Browser-count’** replaces ‘parallel’ and accepts a numeric value as the number of browser to invoke
**‘Partition’** same as before, but will be required if partition-count is specified

 The new commands defaults to running tests with load-balance, while keeping the existing commands to run tests as it does today. The new commands attempt to remove auto-selecting of the number of browsers to invoke which felt convoluted, provide more meaningful naming and simplifies the api usages.  

Sample Usage:
 ember exam --browser-count=10  // browser-count is the single option that controls the number of browsers and will not depend on any other options 
ember exam --partition-count=2  --partition=1  --browser-count=20 // splits the test modules into two partition, running partition 1 against 20 browsers ember exam --partition-count=2 --browser-count=20 // throws an error

Would love to get your feeling on the above @stefanpenner & @rwjblue

README.md
```diff
@@ -30,18 +30,39 @@ $ ember exam --filter='acceptance'
 $ ember exam --server
 ```
 
+A value to an option can be passed with either `=` and a space.
```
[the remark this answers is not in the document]
stefanpenner: ^

README.md
```diff
@@ -30,18 +30,39 @@ $ ember exam --filter='acceptance'
...
+However, to get the unique features of Ember Exam (described in-depth below), you will need to import addon-test-support/load.js to invoke the loadEmberExam function. This is done usually from within test-helper.js but it can be done anywhere that executes prior to your tests loading.
 
 ```js
 // test-helper.js
-import loadEmberExam from 'ember-exam/test-support/load';
+import emberExam from 'ember-exam/test-support/load';
+
+emberExam.loadEmberExam(); // To set an instance of testLoader and pass it to testem
+
+start();
```
stefanpenner: where does start() come from

README.md
```diff
@@ -30,18 +30,39 @@ $ ember exam --filter='acceptance'
...
 
-loadEmberExam();
+
+// emberExam.loadTests(); // To load tests
```
stefanpenner: do we want to leave these comments?

README.md
```diff
@@ -159,6 +180,89 @@ ember exam --split=3 --partition=3
...
+
+To enable `load-balance` feature you will need to import `emberExam` object from `addon-test-support/load.js` then call the two functions shown below within `test-helper.js`
+
+```js
+// test-helper.js
+import emberExam from 'ember-exam/test-support/load';
+
+// An instance of testLoader has to be instantiated before calling `loadTest` and `setupQUnitCallbacks` by calling `loadEmberExam`.
+emberExam.loadEmberExam(); // To set an instance of testLoader and pass it to testem
+emberExam.loadTests(); // To load tests
```
stefanpenner: do we need to diverge from the typical `test-helper.js`? If so, why?

stefanpenner:
left some notes, but this is looking really good.

README.md
choheekim: the start() used to come from ember-qunit to kick off QUnit. After the simplification of ember exam setup (ref. https://github.com/ember-cli/ember-exam/pull/136/commits/207fa49db1c429189887c19c9209c23ece2275fe#diff-412f33e8784be93672251a0e3b3f9ab9R1) it's now only required to call ember-exam's own start function.

README.md
choheekim: updated

README.md
```diff
@@ -159,6 +173,74 @@ ember exam --split=3 --partition=3
...
+Due to the dynamic nature of the `load-balance` option, test file execution order could vary between runs. To reproduce a past test execution follow the next section.
+
+**Note 1** _You must be using ember-cli version 3.2.0 or greater for this feature to work properly._
+**Note 2** _You must be using ember-qunit version 4.1.1 or greater for this feature to work properly._
+**Note 3** _You must be using qunit version 2.8.0 or greater for this feature to work properly._
+**Note 4** _This feature is not currently supported by Mocha._
+
+##### Test Failure Reproduction
+
+Since Ember Exam with `load-balance` loads modules one at a time the order of executing modules is not deterministic. For this reason a test execution with `load-balance` creates a json file to track which browser executes which modules in which order and the file is stored in the root directory and the naming structure is `test-execution-YYYY-MM-DD--HH-MM-SS.json`.
```
stefanpenner: do we need to document usage of `write-execution-file` ?

choheekim: added

README.md
```diff
@@ -27,24 +27,45 @@ Using Ember Exam is fairly straightforward as it extends directly from the defau
...
+
+A value to an option can be passed with either `=` or a space.
+```bash
+# A value of filter is acceptance
+$ ember exam --filter 'acceptance'
+
+# A value of parallel is 2
+$ ember exam --load-balance --parallel=2 --server --no-launch
+
+# A value of parallel is --server
```
step2yeung: I think it would be help to elaborate a bit more here for this case.
```
# If a `=` is not used to pass a value to an option that requires a value, it will take anything passed after a space as it's value
# In this instance, the value of parallel is --server
```

README.md
```diff
@@ -164,6 +175,83 @@ ember exam --split=3 --partition=3
...
+$ ember exam --parallel=<num> --load-balance
+```
+
+`load-balance` allows you to load balance test files against. It will order the test files by test type and load balance the ordered test files between the browsers dynamically rather than statically. `parallel` must be used along with `load-balance` to specify a number of browser(s) and `load-balance` can not be used without `parallel` being specified a number of browser(s).
+
+This option was added to version 1.1 to address execution performance when running against a large test suite.
+
+A way of communication between server and browser to send and receive a test module to load is promise based. Its timeout value is set to be 2 seconds and the timeout can be customized by user by adding `asyncTimeout=[timeout]` in passing url or adding to a testem config.
+
+For example, if you specify load-balance and parallel equals 3, then three browser instances will be created and the output will look something like::
```
step2yeung: lets add the command in a code block

README.md
```diff
@@ -164,6 +175,83 @@ ember exam --split=3 --partition=3
...
+ok 3 Chrome 66.0 - Browser Id 3 - some the other test
+```
+
+You can also specify the `split` and `partition` options with `load-balance` to load a portion of test modules on multiple CI containers.
+
+```bash
+$ ember exam --split=<num> --partition=<num> --parallel=<num> --load-balance
+```
+
+This command will split test files and load-balance tests from the specified partition across the browsers. For example --split=2 -partition=1 --parallel=3 --load-balance, the complete list of test files are split into two halves. With the first half of the list load balanced against three browsers. The output will look something like below:
```
step2yeung: similar here, lets provide a code block with the example command, it would be easier to see

README.md
```diff
@@ -164,6 +175,83 @@ ember exam --split=3 --partition=3
...
+
+```bash
+ok 1 Chrome 66.0 - Exam Partition 1 - browser Id 1 - some test
+ok 2 Chrome 66.0 - Exam Partition 1 - browser Id 2 - another test
+ok 3 Chrome 66.0 - Exam Partition 1 - browser Id 3 - some the other test
+```
+
+Due to the dynamic nature of the `load-balance` option, test file execution order could vary between runs. To reproduce a past test execution the execution must be recorded and passing `--write-execution-file`, in short `--wef` allows generating a JSON file which enables to rerun the past test execution. The option is only allowed when `load-balance` is passed.
```
step2yeung: This paragraph here seems like a repeat with the first paragraph in `test failure reproduction`, do you feel we should combine them?

README.md
```diff
@@ -192,8 +280,8 @@ module.exports = {
 If you are working with [Travis CI](https://travis-ci.org/) then you can also easily set up seeded-random runs based on PR numbers. Similar to the following:
 
 ```js
-const command = [ 'ember', 'exam', '--random' ];
-const pr = process.env.TRAVIS_PULL_REQUEST;
+var command = [ 'ember', 'exam', '--random' ];
```
step2yeung: any reason why we changed these to `var`?

README.md
```diff
@@ -164,6 +176,104 @@ ember exam --split=3 --partition=3
...
+
+```bash
+# The command will read a test execution file specified for `replay-execution` and execute a browser Id(s) from `replay-browser`
+$ ember exam --replay-execution=[string] --replay-browser=[num]
+```
+
+`replay-execution` allows you to specify a path to the json file to run execution against and `replay-browser` is to specify browser ID(s) to execute.
+
+```bash
+# The command will read test-execution-000000.json and load the list of module mapped to browserId 1
```
scalvert:
```bash
# The command will read test-execution-000000.json and load the list of module mapped to browserId 1
```

I'd change this to:

```bash
# The command will read test-execution-000000.json and load the list of modules mapped to browserId 1
```

Note the "s" in modules

README.md
```diff
@@ -164,6 +176,104 @@ ember exam --split=3 --partition=3
...
+$ ember exam --replay-execution=test-execution-000000.json --replay-browser=1
+```
+
+The above command will read `test-execution-000000.json` and load the list of modules which is mapped by browser ID #1.
+
+`replay-browser` can be an array of broweser IDs. For instance `--replay-broweser=1,2` will bring two browsers and execute a list of modules which ran by previous browser #1 and #2.
```
scalvert: > `replay-browser` can be an array of broweser IDs. For instance `--replay-broweser=1,2` will bring two browsers and execute a list of modules which ran by previous browser #1 and #2.

I'd change this to:

`replay-browser` can be an array of browser IDs. For instance `--replay-broweser=1,2` will start two browsers and execute a list of modules which were previously run by browsers #1 and #2.

PR #163, opened 2018-12-03
Upgrades + Updates

stefanpenner:
Upgrades:

* ember-cli
* ember-try
* syntax / es6
* test usage
* test execa usage (soon library itself)

---

Will require a major version bump, as it's babel (6 -> 7) version dropped an old node version support, but beyond that there are no.


stefanpenner:
@rwjblue I'm apparently too stupid to understand how to make travis.yml work correctly, I could use your help:

1. i just manually expanded the matrix, because I couldn't figure out how to make that expansion work correctly
2. I seem to be getting 13 additional test jobs, when I only expect 12

.travis.yml
```diff
@@ -2,38 +2,57 @@
...
+        - yarn test
+
+    - name: "Floating Dependencies"
+      script:
+        - yarn test
+
+    # we recommend new addons test the current and previous LTS
+    # as well as latest stable release (bonus points to beta/canary)
+    - stage: "Additional Tests"
+    - env: EMBER_TRY_SCENARIO=ember-lts-2.16 TEST_FRAMEWORK=ember-qunit
```
rwjblue:
```suggestion
      env: EMBER_TRY_SCENARIO=ember-lts-2.16 TEST_FRAMEWORK=ember-qunit
```

stefanpenner: magik?

stefanpenner: I see this is pretty surprising `.travis.yml` behavior....

```
- stage: "Additional Tests"
- env: FOO=bar
```

is actually 2 jobs:

1. `- stage: "Additional Tests"`
2. `- env: FOO=bar`

xg-wang: the following `- env...` will be other jobs, but belong to same stage as the first 
```
- stage: ''
  env:
```

PR #228, opened 2019-03-26
Renaming and Enabling to read testem yaml file.

choheekim:
1. Rename sharedModuleQueue and browserModuleQueue to replayExecutionModuleQueue and testModuleQueue

2. Add to read yaml file form of testem config


lib/utils/execution-state-manager.js
```diff
@@ -18,10 +18,10 @@ class ExecutionStateManager {
     this._completedBrowsers = 0;
 
     // An array of modules to be load balanced to a number of browsers
-    this._sharedModuleQueue = null;
+    this._testModuleQueue = null;
 
     // A map of browserId to an array of modules to be passed to the browser
```
step2yeung: can we add to the description that this should only be populated when `--replay-execution` is passed?
```
// A map of browserId to an array of test modules. This should only be populated by `--replay-execution`
```

lib/utils/execution-state-manager.js
```diff
@@ -18,10 +18,10 @@ class ExecutionStateManager {
     this._completedBrowsers = 0;
 
     // An array of modules to be load balanced to a number of browsers
```
step2yeung: same here
```
// An array of modules to load balance against browsers. This is used by `--load-balance`
```

node-tests/unit/utils/config-reader-test.js
```diff
@@ -26,4 +26,18 @@ describe('ConfigReader | readTestemConfig', function() {
       'bar'
     );
   });
-});
+
+  it('should read a specified `json` file and return an object read from the file', function() {
+    assert.deepEqual(
+      readTestemConfig('testem.json-file.json', []).foo,
```
step2yeung: Suggestion: all the mock testem config file that are used in this test module.. lets use fixturify to write these files instead of having it in the root of the repo.
```
const cwd = path.resolve('/');  // not sure about this?
  describe('ConfigReader ...', function() {
    beforeEach(function() {
      fixturify.writeSync(cwd, {
        'mock-testem.json': JSON.stringify({
           "foo": "bar"
        })
      });
    });
```

testem.json-file.json
```diff
@@ -0,0 +1,3 @@
+{
```
step2yeung: should be able to remove this after using fixturify

testem.yaml-file.yaml
```diff
@@ -0,0 +1 @@
+foo: bar
```
step2yeung: should be able to remove this after using fixturify

stefanpenner:
released as v2.1.0 🎉

PR #233, opened 2019-03-28
Read broweserId as a string in setReplayExecutionMap and Fix a type

choheekim:
When we record a failed browser id(s) to write it in test-execution-[timestamp].json ember-exam gets a browserId by parsing a url from launcher.setting.test_page. The type of the browserId is string and it's written as a string in the json file. 

For instance, a failedBrowser key in test execution json file will look like:
```
{
  failedBrowser = ["1", "3" ]
}
```
When ember-exam sees `replay-execution` option and read a json file it generates a map which maps from a `string` browserId to a list modules. As the key is stored as a string, when a value for a key is needed to fetch the key should also be string.

### DIFF
--- a/lib/utils/testem-events.js
+++ b/lib/utils/testem-events.js
@@ -38,8 +38,8 @@ class TestemEvents {
     if (browserIdsToReplay && browserIdsToReplay.length > 0) {
       browserIdsToReplay.forEach(browserId => {
         browserModuleMap.set(
-          browserId,
-          executionJson.executionMapping[browserId]
+          browserId.toString(),
+          executionJson.executionMapping[browserId.toString()]
         );
       });
     } else if (executionJson.failedBrowsers.length > 0) {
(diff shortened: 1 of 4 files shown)


lib/utils/testem-events.js
```diff
@@ -38,8 +38,8 @@ class TestemEvents {
     if (browserIdsToReplay && browserIdsToReplay.length > 0) {
       browserIdsToReplay.forEach(browserId => {
         browserModuleMap.set(
-          browserId,
-          executionJson.executionMapping[browserId]
+          browserId.toString(),
```
stefanpenner: what is browserId prior to coercing it into a string here?

choheekim: when we use an option `replay-browser=[num]` the value for the option is number, as in a json file, test-execution.file, it writes browserIds as string it should coerce into a string so that they are using the same type to set and get a module map from the set

choheekim: ^ @stefanpenner

stefanpenner: Thank you

stefanpenner:
released as v2.1.4 🎉

PR #236, opened 2019-03-28
Increase the async iterator timeout from 2s to 15s

step2yeung:
This will give more time for the async iterator to handle events handling.


choheekim:
cool, Thanks!

stefanpenner:
released as v2.1.2 🎉

PR #243, opened 2019-04-04
Update README.md

kategengler:
(Hopefully) increase clarify of setup instructions

- Streamline some sentences
- Fix some typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,22 +43,21 @@ $ ember exam --load-balance --parallel=2 --server --no-launch
 $ ember exam --load-balance --parallel --server --no-launch
 ```
 
-The idea is that you should be able to replace `ember test` with `ember exam` and never look back.
+The idea is that you can replace `ember test` with `ember exam` and never look back.
 
-However, to get the unique features of Ember Exam (described in-depth below), you will need to import `addon-test-support/start.js` to invoke the `start` function. The start function creates an instance of EmberExamTestLoader, loads tests, and invokes start function from either `Ember-Qunit` or `Ember-Mocha`. This is done usually from within `test-helper.js` and should replace the use of `start()` from `Ember-Qunit` or `Ember-Mocha`.
+To get the unique features of Ember Exam (described in-depth below), you will need to **replace** the use of `start()` from `Ember-Qunit` or `Ember-Mocha` in `test-helper.js` with `start()` from `ember-exam`:
 
 ```js
 // test-helper.js
 import start from 'ember-exam/test-support/start';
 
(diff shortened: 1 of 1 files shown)


stefanpenner:
@step2yeung mind doing a review?

README.md
```diff
@@ -43,22 +43,21 @@ $ ember exam --load-balance --parallel=2 --server --no-launch
 $ ember exam --load-balance --parallel --server --no-launch
 ```
 
-The idea is that you should be able to replace `ember test` with `ember exam` and never look back.
+The idea is that you can replace `ember test` with `ember exam` and never look back.
 
-However, to get the unique features of Ember Exam (described in-depth below), you will need to import `addon-test-support/start.js` to invoke the `start` function. The start function creates an instance of EmberExamTestLoader, loads tests, and invokes start function from either `Ember-Qunit` or `Ember-Mocha`. This is done usually from within `test-helper.js` and should replace the use of `start()` from `Ember-Qunit` or `Ember-Mocha`.
+To get the unique features of Ember Exam (described in-depth below), you will need to **replace** the use of `start()` from `Ember-Qunit` or `Ember-Mocha` in `test-helper.js` with `start()` from `ember-exam`:
```
step2yeung: since `start()` from `ember-exam/test-support/start` should replace both `loadEmberExam()` from `ember-exam/test-support/load' and `start()` from `Ember-Qunit` or `Ember-Mocha`. (note, start() from ember-exam will already call loadEmberExam() internally)
Should we call out that `loadEmberExam()` is also no longer needed to be extra clear?

README.md
```diff
@@ -43,22 +43,21 @@ $ ember exam --load-balance --parallel=2 --server --no-launch
...
 
 ```js
 // test-helper.js
 import start from 'ember-exam/test-support/start';
 
-// start() triggers qunit or mocha start after instantiating either qunit or mocha test-loader instance and loading test modules.
+// Options passed to `start` will be passed-through to ember-qunit or ember-mocha
```
choheekim: I think it's important to note that `start` function in ember-exam also instantiate customized test-loader. what do you think? @step2yeung

step2yeung: @choheekim Question back to you:
In the context of `start()`, do you feel users will need to know about the customized test loader?

I feel if we have a v3 upgrade doc, which describes the differences from v2 to v3, then it would be good to have this for context. From a readme's perspective, I am not sure if users will benefit from knowing this.

kategengler: As a user, I don't care how it works, so long as it does 😝

choheekim: sounds reasonable :)

stefanpenner:
thanks @kategengler!

Issue #238, opened 2019-04-01
2.1.x does not work with latest stable ember-mocha release


kevinansfield:
I've added the `start()` call from `ember-exam` to our `test-helper.js` file (even though the current ember-mocha release doesn't use an explicit `start()` call) which allows our tests to run but I still get this error when attempting to run our test suite and parallelization doesn't occur:

```
Global error: Uncaught Error: Unable to find a suitable test loader. You should ensure that one of `ember-qunit` or `ember-mocha` are added as dependencies.
```

Looking at the [`getTestLoader()` code here](https://github.com/ember-cli/ember-exam/blob/68641d6f8ce5a2208b8a8d5d0336823aa18706ea/addon-test-support/-private/get-test-loader.js#L13-L15) it's checking for the existence of `ember-mocha/test-loader` but `ember-mocha` does not have such an export or file.

ember-exam@2.1.1
ember-mocha@0.14.0

step2yeung:
@kevinansfield Thanks for filing this issue!
It looks like you were getting this error with `TryGhost/Ghost-Admin` ?

step2yeung:
I was able to reproduce this. The problem is due to test-loader is available on master, but not in the releases yet. Will be addressing this

amk221:
I don't use ember-mocha, but might I be experiencing the same issue?

```
not ok 1 Chrome 73.0 - [undefined ms] - Global error: Uncaught TypeError: (0 , _load.default) is not a function at http://localhost:7357/assets/tests-dev.js, line 48328
    ---
        Log: |
            { type: 'error',
              testContext: {},
              text:
               'Uncaught TypeError: (0 , _load.default) is not a function at http://localhost:7357/assets/tests-dev.js, line 48328\n' }
    ...
```

step2yeung:
@amk221 `ember-exam 2.1.2` removed the load API and instead introduces the `start()` API
Please see the Read me for how to set this up.

amk221:
Oh, sorry - I _did_ check the changelog, but it's not mentioned.

amk221:
...wait also, I'm already using that start helper

step2yeung:
@amk221 no worries, what you faced is reasonable since it was a breaking change that was introduced into a minor version bump. #242 will fix this. So you can use `load()` instead.

stefanpenner:
https://github.com/ember-cli/ember-exam/pull/242 Was merged and released.

maprules1000:
[...]

PR #259, opened 2019-04-12
[wip] Make mocha work with ember-exam v3

step2yeung:
**Problem:**
There is no `test-loader` in the latest release of ember-mocha. Hence users are seeing the following error in the browser console error:
```
Uncaught Error: Unable to find a suitable test loader. You should ensure that 
one of `ember-qunit` or `ember-mocha` are added as dependencies. at 
http://localhost:7357/assets/test-support.js, line 51528
```
reference: #238 

**Fix:**
- Change `getTestLoader()` to search for `ember-mocha` or `ember-qunit` from `require`
- Change `ember-exam-mocha-test-loader` to extend from `ember-cli-test-loader` instead of `ember-mocha/test-loader` which has not been released yet

**Other work:**
- move `loadTests` back to `load.js`

**Tests:**
- [x] Test with Ghost-admin
- [ ] Add test that tests all ember-exam features against mocha

### DIFF
--- a/addon-test-support/-private/ember-exam-mocha-test-loader.js
+++ b/addon-test-support/-private/ember-exam-mocha-test-loader.js
@@ -1,10 +1,12 @@
 import getUrlParams from './get-url-params';
 import splitTestModules from './split-test-modules';
-import { TestLoader } from 'ember-mocha/test-loader';
+import { describe, it } from 'mocha';
+
+// get testLoader from 'ember-qunit/test-loader' once a new version get's released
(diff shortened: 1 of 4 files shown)


step2yeung:
This won't work since tests are being ran twice, one by calling start() here, another time by autostart from ember-mocha. The only way this will work is to re-introduce monkey-patching of ember-cli-test-loader, which we moved away from in v3.
Mitigation is to use < v3.0 of ember-exam

PR #265, opened 2019-04-22
Generate module detail json file and enable to log out a module queue statue when a browser exit.

choheekim:
This change adds
1. to generate module detail json file when running in load-balance
2. to log a module queue status when a browser exit when in load-balance mode

A load-balance mode works the best when each module duration takes a fair amount of time. For now we do not have any threshold or blocker when a long run module comes in. Having module detail json file is a first step to help setting threshold or something to make load-balance the best when comes in module duration. Additionally the module detail json file can also help investigating if any module causes a whole test suite execution last long.

Currently after test suite execution in load balance mode we do not have any information of how the test goes. Sometime we've observed that a last browser runs multiple modules (two or three) which we do not expect. Logging a length of the test module queue when a browser exits can help a bit to investigate the issue. In future I will also add a list of modules left when a browser exits to provide more informative logging.

### DIFF
--- a/addon-test-support/-private/ember-exam-qunit-test-loader.js
+++ b/addon-test-support/-private/ember-exam-qunit-test-loader.js
@@ -69,6 +69,7 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
     }
 
     super.loadModules();
+    this.setupModuleMetadataHandler();
 
     if (modulePath || filePath) {
       this._testModules = filterTestModules(
@@ -119,6 +120,17 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
   }
 
   /**
+   * setupModuleMetadataHandler() register QUnit callback to enable generating module metadata file.
+   */
+  setupModuleMetadataHandler() {
+    this._qunit.moduleDone((details) => {
+      // testem:module-done-metadata is sent to server to keep track of test module details.
+      // 'details' contains module metadata like module name, total number of assertion ran in the module, and module runtime.
+      this._testem.emit('testem:module-done-metadata', details);
+    });
+  }
+
+  /**
    * setupLoadBalanceHandlers() registers QUnit callbacks needed for the load-balance option.
    */
   setupLoadBalanceHandlers() {
@@ -166,10 +178,7 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
       return nextModuleHandler();
     });
 
-    this._qunit.moduleDone((details) => {
-      // testem:module-done-detail is sent to server to keep track of test module details.
-      // 'details' contains module metadata like module name, total number of assertion ran in the module, and module runtime.
-      this._testem.emit('testem:module-done-detail', details);
+    this._qunit.moduleDone(() => {
       return nextModuleHandler();
     });
   }
--- a/addon-test-support/-private/ember-exam-qunit-test-loader.js
+++ b/addon-test-support/-private/ember-exam-qunit-test-loader.js
@@ -167,7 +167,8 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
     });
 
     this._qunit.moduleDone((details) => {
-      // testem:module-done-detail is sent to server to keep tracking of test module details.
+      // testem:module-done-detail is sent to server to keep track of test module details.
+      // 'details' contains module metadata like module name, total number of assertion ran in the module, and module runtime.
       this._testem.emit('testem:module-done-detail', details);
       return nextModuleHandler();
     });
--- a/addon-test-support/-private/ember-exam-qunit-test-loader.js
+++ b/addon-test-support/-private/ember-exam-qunit-test-loader.js
@@ -123,10 +123,10 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
    * setupModuleMetadataHandler() register QUnit callback to enable generating module metadata file.
    */
   setupModuleMetadataHandler() {
-    this._qunit.moduleDone((details) => {
+    this._qunit.moduleDone((metadata) => {
       // testem:module-done-metadata is sent to server to keep track of test module details.
-      // 'details' contains module metadata like module name, total number of assertion ran in the module, and module runtime.
-      this._testem.emit('testem:module-done-metadata', details);
+      // 'metadata' contains module name, total number of assertion ran in the module, and module runtime.
+      this._testem.emit('testem:module-done-metadata', metadata);
     });
   }
 
--- a/addon-test-support/-private/ember-exam-qunit-test-loader.js
+++ b/addon-test-support/-private/ember-exam-qunit-test-loader.js
@@ -166,7 +166,9 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
       return nextModuleHandler();
     });
 
-    this._qunit.moduleDone(() => {
+    this._qunit.moduleDone((details) => {
+      // testem:module-done-detail is sent to server to keep tracking of test module details.
+      this._testem.emit('testem:module-done-detail', details);
       return nextModuleHandler();
     });
   }
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -248,7 +248,11 @@ module.exports = TestCommand.extend({
     let additionalEvents = this._setupAndGetBrowserSocketEvents(config);
 
     if (commandOptions.loadBalance || commandOptions.replayExecution) {
-      const loadBalancingEvents = this._getLoadBalancingBrowserSocketEvents(this.commands, this.testemEvents);
+      const loadBalancingEvents = this._getLoadBalancingBrowserSocketEvents({
+        isLoadBalance: this.commands.get('loadBalance'),
+        isReplayExecution: this.commands.get('replayExecution'),
+        isWriteExecutionFile: this.commands.get('writeExecutionFile')
+      }, this.testemEvents );
       additionalEvents = Object.assign(additionalEvents, loadBalancingEvents);
     }
 
@@ -390,25 +394,25 @@ module.exports = TestCommand.extend({
    * @param {Object} commands
    * @param {Object} testemEvents
    */
-  _getLoadBalancingBrowserSocketEvents(commands, testemEvents) {
+  _getLoadBalancingBrowserSocketEvents({ isLoadBalance, isReplayExecution, isWriteExecutionFile } , testemEvents) {
     const events = {};
     events['testem:set-modules-queue'] = function(modules, browserId) {
       testemEvents.setModuleQueue(
         browserId,
         modules,
-        commands.get('loadBalance'),
-        commands.get('replayExecution')
+        isLoadBalance,
+        isReplayExecution
       );
     };
     events['testem:next-module-request'] = function(browserId) {
       testemEvents.nextModuleResponse(
         browserId,
         this.socket,
-        commands.get('writeExecutionFile')
+        isWriteExecutionFile
       );
     };
     events['test-result'] = function(result) {
-      if (result.failed && commands.get('writeExecutionFile')) {
+      if (result.failed && isWriteExecutionFile) {
         const browserId = getBrowserId(this.launcher.settings.test_page);
         testemEvents.recordFailedBrowserId(browserId);
       }
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { addToQuery } = require('../utils/query-helper');
+const log = require('npmlog'); // eslint-disable-line node/no-extraneous-require
 const {
   combineOptionValueIntoArray,
   getBrowserId,
@@ -10,6 +11,7 @@ const TestemEvents = require('../utils/testem-events');
 const TestCommand = require('ember-cli/lib/commands/test'); // eslint-disable-line node/no-unpublished-require
 const TestServerTask = require('./task/test-server');
 const TestTask = require('./task/test');
+
 
 module.exports = TestCommand.extend({
   name: 'exam',
@@ -277,7 +279,7 @@ module.exports = TestCommand.extend({
         browserId,
         ui,
         commands.get('loadBalance'),
-        testExecutionFileName,
+        Date.now(),
         commands.get('writeExecutionFile')
       );
     };
@@ -325,6 +327,9 @@ module.exports = TestCommand.extend({
         this.socket,
         commands.get('writeExecutionFile')
       );
+    };
+    events['testem:module-done-detail'] = function(details) {
+      testemEvents.recordModuleRunDetail(details.name, details.total, details.runtime);
     };
     events['test-result'] = function(result) {
       if (result.failed && commands.get('writeExecutionFile')) {
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -278,17 +278,40 @@ module.exports = TestCommand.extend({
     const commands = this.commands;
     const testemEvents = this.testemEvents;
     const ui = this.ui;
-    const events = config.custom_browser_socket_events || {};
-    const testExecutionFileName = `test-execution-${Date.now()}.json`;
-    const browserCount = Object.keys(config.testPage).length;
-    const browserExitHandler = function() {
-      const browserId = getBrowserId(this.launcher.settings.test_page);
+
+    const browserExitHandler = function(failed = false) {
+
+      const launcherId = this.launcher.id;
+      if (!failed && commands.get('loadBalance')) {
+        try {
+          const browserId = getBrowserId(this.launcher.settings.test_page);
+          log.info(`Browser ${browserId} exiting. [ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]`);
+        } catch (err) {
+          const message = testemEvents.stateManager.getTestModuleQueue() === null ?
+          'testModuleQueue is not set.' :
+          `[ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]`;
+          if (typeof err === 'object' && err !== null) {
+            err.message = `${err.message} \n ${message}`;
+          } else {
+            throw new Error(message);
+          }
+        }
+      }
+
+      // config.testPage is undefined when parallization options are not used
+      // Set browserCount default value to 1 to allow
+      let browserCount = 1;
+      // When using multiple browsers config.testPage is an array of test page urls.
+      if (typeof config.testPage !== 'undefined') {
+        browserCount = Object.keys(config.testPage).length;
+      }
 
       testemEvents.completedBrowsersHandler(
         browserCount,
-        browserId,
+        launcherId,
         ui,
-        commands
+        commands,
+        Date.now()
       );
     };
 
@@ -310,7 +333,7 @@ module.exports = TestCommand.extend({
         testemEvents.recordFailedBrowserId(browserId);
       }
 
-      browserExitHandler.call(this);
+      browserExitHandler.call(this, true);
     };
 
     return this._getModuleMetadataAndBrowserExitSocketEvents(browserExitHandler, browserFailureHandler);
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -1,7 +1,8 @@
 'use strict';
 
 const { addToQuery } = require('../utils/query-helper');
-const log = require('npmlog'); // eslint-disable-line node/no-extraneous-require
+// npmlog is used to write to testem server logs and `--testem-debug` enables to save the log file.
+const log = require('npmlog');
 const {
   combineOptionValueIntoArray,
   getBrowserId,
@@ -329,7 +330,13 @@ module.exports = TestCommand.extend({
       );
     };
     events['testem:module-done-detail'] = function(details) {
-      testemEvents.recordModuleRunDetail(details.name, details.total, details.runtime);
+      testemEvents.recordModuleMetaData(
+        {
+          name: details.name,
+          total: details.total,
+          runtime: details.runtime
+        }
+      );
     };
     events['test-result'] = function(result) {
       if (result.failed && commands.get('writeExecutionFile')) {
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -245,7 +245,15 @@ module.exports = TestCommand.extend({
    */
   _generateCustomConfigs(commandOptions) {
     const config = this._super._generateCustomConfigs.apply(this, arguments);
-    config.custom_browser_socket_events = this._addBrowserSocketEvents(commandOptions, config);
+    const events = config.custom_browser_socket_events || {};
+
+    this._addBrowserSocketEvents(config, events);
+
+    if (commandOptions.loadBalance || commandOptions.replayExecution) {
+      this._addLoadBalancingBrowserSocketEvents(events, this.commands, this.testemEvents);
+    }
+
+    config.custom_browser_socket_events = events;
 
     if (!commandOptions.loadBalance && !commandOptions.replayExecution && !commandOptions.parallel )
       return config;
@@ -263,13 +271,12 @@ module.exports = TestCommand.extend({
   },
 
   /**
-   * Returns an event object that enables to send and receive module metadata and load-balance if needed
+   * Add events to enable to send and receive module metadata
    *
-   * @param {*} commandOptions
-   * @param {*} config
+   * @param {Object} config
+   * @param {Object} events
    */
-  _addBrowserSocketEvents(commandOptions, config) {
-    const events = config.custom_browser_socket_events || {};
+  _addBrowserSocketEvents(config, events) {
     const commands = this.commands;
     const testemEvents = this.testemEvents;
     const ui = this.ui;
@@ -309,6 +316,17 @@ module.exports = TestCommand.extend({
       browserExitHandler.call(this);
     };
 
+    this._addCustomBrowserSocketEvents(events, browserExitHandler, browserFailureHandler);
+  },
+
+  /**
+   * Add browser socket events are needed for both with load-balance and without load-balance
+   *
+   * @param {Object} events
+   * @param {Object} browserExitHandler
+   * @param {Object} browserFailureHandler
+   */
+  _addCustomBrowserSocketEvents(events, browserExitHandler, browserFailureHandler) {
     let init = false;
 
     events['tests-start'] = function() {
@@ -318,6 +336,9 @@ module.exports = TestCommand.extend({
         init = true;
       }
     };
+
+    events['after-tests-complete'] = browserExitHandler;
+    events['disconnect'] = browserFailureHandler;
 
     events['testem:module-done-metadata'] = (details) => {
       // Ensure module detail is available
@@ -331,19 +352,6 @@ module.exports = TestCommand.extend({
         );
       }
     };
-
-    events['after-tests-complete'] = browserExitHandler;
-    events['disconnect'] = browserFailureHandler;
-
-    if (commandOptions.loadBalance || commandOptions.replayExecution) {
-      this._addLoadBalancingBrowserSocketEvents(
-        events,
-        commands,
-        testemEvents
-      );
-    }
-
-    return events;
   },
 
   /**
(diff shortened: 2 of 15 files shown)


choheekim:
Currently ci is failing and I'm working on fix and also adding more tests

choheekim:
ci's fixed. Will be adding more tests

lib/commands/exam.js
```diff
@@ -285,6 +299,9 @@ module.exports = TestCommand.extend({
         commands.get('writeExecutionFile')
       );
     };
+    events['testem:module-done-detail'] = function(details) {
+      testemEvents.recordModuleRunDetail(details.name, details.total, details.runtime);
```
step2yeung: what if we simplified this to 
```
// notice I also renamed recordModuleRunDetail to recordModuleMetadata
testemEvents.recordModuleMetadata(   
  {
    details.name,
    details.total,
    details.runtime
  }
);
```

Then everywhere we can simplify the params to:
```
recordModuleMetadata(metadata)
addMetadataToModuleMetadata(metadata)
```

lib/utils/execution-state-manager.js
```diff
@@ -149,6 +152,15 @@ class ExecutionStateManager {
     return this._browserToModuleMap;
   }
 
+  /**
+   * Returns an array of modules run details
+   *
+   * @returns {Array}
+   */
+  getModuleRunDetails() {
```
step2yeung: suggestion: getModulesMetadata

lib/utils/execution-state-manager.js
```diff
@@ -165,6 +177,17 @@ class ExecutionStateManager {
...
   }
 
+  /**
+   * Pushes the module detail into the moduleRunDetails array
+   *
+   * @param {*} moduleName
+   * @param {*} totalAssertionNum
+   * @param {*} moduleTotalDuration
+   */
+  addModuleDoneDetailToModuleRunDetails(moduleName, totalAssertionNum, moduleTotalDuration) {
```
step2yeung: as suggested above, could be renamed & simplified to `addMetadataToModuleMetadata(metadata){`

lib/utils/testem-events.js
```diff
@@ -121,6 +121,17 @@ class TestemEvents {
...
   }
 
+  /**
+   * Record the module run details to the stateManager
+   *
+   * @param {*} moduleName
+   * @param {*} totalAssertionNum
+   * @param {*} moduleTotalDuration
+   */
+  recordModuleRunDetail(moduleName, totalAssertionNum, moduleTotalDuration) {
```
step2yeung: as suggested above, could be renamed & simplified to recordModuleMetadata(metadata){

lib/utils/testem-events.js
```diff
@@ -152,31 +163,49 @@ class TestemEvents {
...
   }
 
+  /**
+   * Creates a file with targetJsonObject
+   *
+   * @param {*} fileName
+   * @param {*} targetJsonObject
+   * @param {*} option
+   */
+  _generatesJsonExecutionFile(fileName, targetJsonObject, option = {}) {
```
step2yeung: suggestion: rename to `_writeJsonToFile`

lib/utils/testem-events.js
```diff
@@ -152,31 +163,49 @@ class TestemEvents {
...
    * @param {string} writeExecutionFile
    */
-  completedBrowsersHandler(browserCount, ui, loadBalance, fileName, writeExecutionFile) {
+  completedBrowsersHandler(browserCount, ui, loadBalance, currentDate, writeExecutionFile) {
     let browsersCompleted = false;
 
     this.stateManager.incrementCompletedBrowsers();
     if (this.stateManager.getCompletedBrowser() === browserCount) {
+      const moduleDetailFileName = path.join(this.root, `module-run-details-${currentDate}.json`);
+      const sortedModuleRunDetails = this.stateManager.getModuleRunDetails().sort((a, b) => { return b.totalDuration - a.totalDuration; });
```
step2yeung: can we hide this behind a flag? It seems more like a power user feature

lib/commands/exam.js
```diff
@@ -239,14 +241,26 @@ module.exports = TestCommand.extend({
     const testemEvents = this.testemEvents;
     const ui = this.ui;
     const events = config.custom_browser_socket_events || {};
-    const testExecutionFileName = `test-execution-${Date.now()}.json`;
-    const browserExitHandler = function() {
+
+    const browserExitHandler = function(failed = false) {
+      if (!failed && commands.get('loadBalance')) {
+        let browserId;
+        try {
```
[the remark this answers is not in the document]
choheekim: 1. I think just because it's not able to find a browser id to log out it shouldn't stop executing the test..
2. Now that I think about it..hmm I guess there wouldn't be no cases of browserId not being available?
3. I will change to log out something like `A browser Id is not available. A browser is exiting..`

lib/utils/execution-state-manager.js
```diff
@@ -13,6 +13,9 @@ class ExecutionStateManager {
     // A map of browerId to test modules executed for the current test execution.
     this._browserToModuleMap = new Map();
 
+    // An array keeping the module execution details
+    this._moduleRunDetails = [];
```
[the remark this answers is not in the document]
arthirm: Or `moduleExecutionMetadata`. Am assuming its to record module execution details.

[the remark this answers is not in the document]
step2yeung: adding `execution` seems redundant since the state manager is already a class to store execution state

lib/commands/exam.js
```diff
@@ -239,14 +242,24 @@ module.exports = TestCommand.extend({
...
+          browserId = getBrowserId(this.launcher.settings.test_page);
+          log.info(`A browser ${browserId} exiting... [ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]`);
+        } catch(err) {
+          throw new Error(`A browser Id is not available. A browser is exiting.. [ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]`);
```
step2yeung: Can we add the message here to the error to be thrown so we don't lose the actual error that's thrown?

lib/utils/execution-state-manager.js
```diff
@@ -165,6 +177,15 @@ class ExecutionStateManager {
     this._browserToModuleMap.set(browserId, browserModuleList);
   }
 
+  /**
+   * Pushes the module detail into the moduleMetadata array
+   *
+   * @param {*} metaData
+   */
+  addMetadataToModuleMetadata(metaData) {
+    return this._moduleMetadata.push({name: metaData.name, total: metaData.total, runtime: metaData.runtime});
```
step2yeung: i think we can just do `this._moduleMetadata.push(metaData);` here?

lib/utils/testem-events.js
```diff
@@ -152,31 +161,56 @@ class TestemEvents {
...
-  completedBrowsersHandler(browserCount, ui, loadBalance, fileName, writeExecutionFile) {
+  completedBrowsersHandler(browserCount, ui, loadBalance, currentDate, writeExecutionFile) {
     let browsersCompleted = false;
 
     this.stateManager.incrementCompletedBrowsers();
     if (this.stateManager.getCompletedBrowser() === browserCount) {
+      const moduleDetailFileName = path.join(this.root, `module-run-details-${currentDate}.json`);
+      const sortedModuleMetadata = this._getSortedModuleMetaData(this.stateManager.getModuleMetadata());
+
+      this._writeJsonToFile(moduleDetailFileName, sortedModuleMetadata, { spaces: 2});
```
step2yeung: nitpick: spacing `{ spaces: 2 });`

lib/utils/testem-events.js
```diff
@@ -152,31 +161,56 @@ class TestemEvents {
...
+   * @param {string} currentDate
    * @param {string} writeExecutionFile
    */
-  completedBrowsersHandler(browserCount, ui, loadBalance, fileName, writeExecutionFile) {
+  completedBrowsersHandler(browserCount, ui, loadBalance, currentDate, writeExecutionFile) {
     let browsersCompleted = false;
 
     this.stateManager.incrementCompletedBrowsers();
     if (this.stateManager.getCompletedBrowser() === browserCount) {
+      const moduleDetailFileName = path.join(this.root, `module-run-details-${currentDate}.json`);
```
step2yeung: Don't see a flag here for `module-run-details-.json`
Is the plan to write to disk for every run?

choheekim: I thought that the module metatdata file will be useful for every runs, but open to any thoughts :)

step2yeung: I had the following scenario in mind:
- if the test suite is pretty well optimize, what would the user gain from this file? They may want to say I no longer need this file but would have no way to turn it off.

Side note: I'm sure this will come in future revision, documentation on what the file means and how it can be used.

lib/commands/exam.js
```diff
@@ -239,14 +242,24 @@ module.exports = TestCommand.extend({
...
-    const testExecutionFileName = `test-execution-${Date.now()}.json`;
-    const browserExitHandler = function() {
+
+    const browserExitHandler = function(failed = false) {
+      if (!failed && commands.get('loadBalance')) {
+        try {
+          const browserId = getBrowserId(this.launcher.settings.test_page);
+          log.info(`Browser ${browserId} exiting. [ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]`);
+        } catch(err) {
+          const errMsg = `A browser Id is not available: ${err.message}`;
```
stefanpenner: Several notes about re-throwing errors with additional data.

1. `err` from catch can by anything, null, undefined, a string, etc, so we should check that before accessing `.message`
2. the method of error wrapping may loose important information, let me instead recommend we simply augment the existing message.

```js

const message = `[ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]` 
 if (typeof err === 'object' && err !== null) {
  // the err thrown can have its message augmented, let's augment it, but leave the original error as it. This ensures we do not lose important error context.
  err.message = `${err.message} \n ${message}`;
} else {
  // we are in a catch, but the `err` doesn't appear to be a real error and since we do want to include our message, we create a new error with said message and throw it.
  throw new Error(message);
}

throw err;
```

lib/utils/testem-events.js
```diff
@@ -152,31 +161,56 @@ class TestemEvents {
...
+    } catch (err) {
+      err.file =  err.file || fileName;
+      throw err;
+    }
+  }
+
+  /**
+   * Returns a sorted module metadata from statemanager.
+   */
+  _getSortedModuleMetaData(moduleMetadata) {
```
stefanpenner: nit:  this appears to be a helper method, not reliant on `this` nor to be overridden by sub-classes. We should extract this from the prototype, and instead make it a module-level "pure" function.

lib/utils/testem-events.js
```diff
@@ -152,31 +161,56 @@ class TestemEvents {
...
   }
 
+  /**
+   * Creates a file with targetJsonObject
+   *
+   * @param {*} fileName
+   * @param {*} targetJsonObject
+   * @param {*} option
+   */
+  _writeJsonToFile(fileName, targetJsonObject, option = {}) {
```
stefanpenner: This looks like a utility function, which does not depend on `this`, nor do we want/expect subclasses to override it. Let's move this off the prototype, and instead make it a top-level function in this module, as having it on the prototype is misleading (it suggests it depends on some internal instance state)

lib/utils/testem-events.js
```diff
@@ -152,31 +161,56 @@ class TestemEvents {
...
+    try {
+      fs.writeJsonSync(fileName, targetJsonObject, option);
+    } catch (err) {
+      err.file =  err.file || fileName;
```
stefanpenner: the value provided to the catch clause in JavaScript is funky, it can literally be anything. To avoid our catch clause itself crashing, I recommend that we are more defensive here, and ensure that the error is in-fact an object who can reasonably have properties set on it.

```
if (typeof err === 'object' && err !== null ) {
    err.file = err.file || fileName;
}

throw err;
```

lib/utils/testem-events.js
```diff
@@ -152,31 +161,56 @@ class TestemEvents {
...
+      throw err;
+    }
+  }
+
+  /**
+   * Returns a sorted module metadata from statemanager.
+   */
+  _getSortedModuleMetaData(moduleMetadata) {
+    return moduleMetadata.sort((a, b) => { return b.runtime - a.runtime; });
```
stefanpenner: nit: `=>` functions can drop the `{` and `return` 

```js
sort((a, b)  => b.runtime - a.runtime);
```

lib/utils/testem-events.js
```diff
@@ -121,6 +121,15 @@ class TestemEvents {
     }
   }
 
+  /**
+   * Record the module run details to the stateManager
+   *
+   * @param {*} metaData
+   */
+  recordModuleMetaData(metaData) {
+    this.stateManager.addMetadataToModuleMetadata(metaData);
```
stefanpenner: After reading this PR, is there a reason for us to store this metaData on the stateManager, or simply on this test-events object. 

My thoughts:

1) it seems to only be used by this object, shuttling the values to the second object seems redundant
2) it doesn't seems related to the state management itself, largely just a way of serializing data er: event meta data.

stefanpenner:
I like the idea of having these meta-data available!

lib/utils/testem-events.js
choheekim: I agree.. If we are making to have module metadata file as optional I'm thinking shouldn't it be an optional for every cases whether or not it's using load-balance? Even without using load-balance I think the metadata can be somewhat useful for many cases..

step2yeung: `Even without using load-balance I think the metadata can be somewhat useful for many cases.` Agreed! Although we don't have any metadata info for non-load balance runs atm.

choheekim: Glad you agreed! I think we still can use QUnit callbacks to make metadata available for both non-load balance and load balance.

choheekim:
Updates:
- Working on enabling metadata for both non-load balance and load-balance. Currently qunit.done callback and customized testem event handlers are only available for load-balance feature. As sending and receiving module details over browser and server is achieve by events, callbacks and event handlers should also be available for non-load balance.

lib/utils/testem-events.js
choheekim: My thought on storing the metaData on the stateManager is due to
- I consider the module detail as a part of test state
- and module details do not really seem to belong to testem-event where testem-event object is to coordinate testem events.. 

but open to your thoughts

addon-test-support/-private/ember-exam-qunit-test-loader.js
```diff
@@ -107,6 +108,17 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
     super.unsee(moduleName);
   }
 
+  /**
+   * setupModuleMetadataHandler() register QUnit callback to enable generating module metadata detail file.
```
step2yeung: nitpick: `setupModuleMetadataHandler() register QUnit callback to enable generating module metadata file.`

addon-test-support/-private/ember-exam-qunit-test-loader.js
```diff
@@ -107,6 +108,17 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
...
+   */
+  setupModuleMetadataHandler() {
+    this._qunit.moduleDone((details) => {
+      // testem:module-done-detail is sent to server to keep track of test module details.
+      // 'details' contains module metadata like module name, total number of assertion ran in the module, and module runtime.
```
step2yeung: should we just name this as `metadata` ?

choheekim: yes I think we should :)

node-tests/acceptance/exam-test.js
```diff
@@ -414,7 +441,7 @@ describe('Acceptance | Exam Command', function() {
           'failure-dist',
           '--load-balance',
           '--parallel',
-          '2',
+          '1',
```
step2yeung: curious, whats the reason for this change?

lib/commands/exam.js
```diff
@@ -206,50 +217,66 @@ module.exports = TestCommand.extend({
...
+      }
+
       testemEvents.completedBrowsersHandler(
         browserCount,
         ui,
-        commands.get('loadBalance'),
-        testExecutionFileName,
-        commands.get('writeExecutionFile')
+        commands,
+        Date.now()
```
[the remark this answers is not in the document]
choheekim: The original reason to have Date.now() as a param is for testing purpose. e.g. https://github.com/ember-cli/ember-exam/pull/265/files#diff-b9793fc80c1674089c42a19b160ee155R210

but there might be a way to test it against beside the way it's doing.. let me explore more on this

node-tests/acceptance/exam-test.js
choheekim: I've made changes because it seems to create multiple test-execution file from multiple events are called and `browserExitHandler` is executed multiple times. After I remove `disconnect` callback registering I made changes back to test against 2 browsers.

stefanpenner:
Is this ready to be reviewed? If not, what is remaining?

choheekim:
@stefanpenner this's ready to be reviewed. I'm working on fixing failing tests in CI but I think the fix should be subtle one

lib/commands/exam.js
```diff
@@ -206,50 +217,75 @@ module.exports = TestCommand.extend({
    */
   _generateCustomConfigs(commandOptions) {
     const config = this._super._generateCustomConfigs.apply(this, arguments);
+    const events = config.custom_browser_socket_events || {};
+
+    this._addBrowserSocketEvents(config, events);
```
step2yeung: instead of modifying an input, can we have a pure function that returns an object with the events?
```
_generateCustomConfigs(commandOptions) {
    const config = this._super._generateCustomConfigs.apply(this, arguments);
    let additionalEvents =  this._getBrowserSocketEvents(config);

    if (commandOptions.loadBalance || commandOptions.replayExecution) {
      const loadBalancingEvents = this._getLoadBalancingBrowserSocketEvents(this.commands, this.testemEvents);
      additionalEvents = { ...additionalEvents, ...loadBalancingEvents };
    }

    config.custom_browser_socket_events = { ...config.custom_browser_socket_events, ...additionalEvents };
```

lib/commands/exam.js
```diff
@@ -258,9 +294,20 @@ module.exports = TestCommand.extend({
...
+  },
+
+  /**
+   * Add browser socket events are needed for both with load-balance and without load-balance
+   *
+   * @param {Object} events
+   * @param {Object} browserExitHandler
+   * @param {Object} browserFailureHandler
+   */
+  _addCustomBrowserSocketEvents(events, browserExitHandler, browserFailureHandler) {
```
step2yeung: `_addBrowserSocketEvents` and `_addCustomBrowserSocketEvents` sound a bit too similar... it make me think, what is the difference?
could we rename this to `_getModuleMetadataAndBrowserExitSocketEvents` ?

lib/commands/exam.js
```diff
@@ -206,50 +217,75 @@ module.exports = TestCommand.extend({
...
-   * Returns a object of event handlers which enables load balancing.
-   * These event handlers will be registered on Testem's browserTestRunner socket instance
+   * Add events to enable to send and receive module metadata
    *
    * @param {Object} config
-   * @return {Object} events
+   * @param {Object} events
    */
-  _addLoadBalancingBrowserSocketEvents(config) {
+  _addBrowserSocketEvents(config, events) {
```
step2yeung: Could we rename this to `_setupAndGetBrowserSocketEvents`?
I feel this reflects the code more.

lib/commands/exam.js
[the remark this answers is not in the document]
step2yeung: got it. How about having currentDate as a default parameter?
```
completedBrowsersHandler(browserCount, ui, commands, currentDate = Date.now()) {
```

lib/commands/exam.js
```diff
@@ -270,6 +317,42 @@ module.exports = TestCommand.extend({
...
+    events['disconnect'] = function() {
+      // when test runs without launcher (with --no-launch) process object is not avaialble.
+      // To prevent handling exiting browser browser disconnects from errors `disconnect` callback's needed to be registered.
+      if (this.process === null) {
+        browserFailureHandler.bind(this)();
+      }
+    }
+
+    //register disconnect when no-launch is availble
+
```
step2yeung: nitpick: remove blank line

node-tests/acceptance/exam-test.js
```diff
@@ -293,6 +294,18 @@ describe('Acceptance | Exam Command', function() {
...
+    function assertModuleDetailJson(output) {
+      let moduleRunDetailJsonPath = path.join(
+        process.cwd(),
+        output.match(/module-metadata-([0-9]*).json/g)[0]
+      );
+      assert.ok(
+        fs.existsSync(moduleRunDetailJsonPath),
+        'module run detail json written to root'
+      );
+      unlinkFiles.push(moduleRunDetailJsonPath);
```
step2yeung: if the assert fails.. this line won't get executed? so it may be better to move this above the assert.

choheekim:
CI's failing. I've been working on other task, so I will be working on fix for failing test this weekend & next week.

choheekim:
@stefanpenner @step2yeung CI's passing after landing a pr to use node 8 in ember-exam :) Ready to be merged

stefanpenner:
:shipit:

step2yeung:
Just one minor comment, then we should be good to merge!

choheekim:
Addressed the last comment and rebased :)

choheekim:
@arthirm @step2yeung @stefanpenner Could you guys give another round of review as now CI's fixed :) ?

lib/commands/exam.js
```diff
@@ -299,6 +348,42 @@ module.exports = TestCommand.extend({
...
+  },
+
+  /**
+   * Return an event object which enables load balancing.
+   * These event handlers will be registered on Testem's browserTestRunner socket instance
+   *
+   * @param {Object} commands
+   * @param {Object} testemEvents
+   */
+  _getLoadBalancingBrowserSocketEvents(commands, testemEvents) {
```
step2yeung: instead of passing all of `this.commands`.. can we do 
```
this._getLoadBalancingBrowserSocketEvents({
  this.commands.get('loadBalance'),
  this.commands.get('replayExecution'),
  this.commands.get('writeExecutionFile')
}, this.testemEvents );
...
_getLoadBalancingBrowserSocketEvents({ isLoadBalance, isReplayExecution, isWriteExecutionFile} , testemEvents) {
```

lib/utils/execution-state-manager.js
```diff
@@ -165,6 +177,15 @@ class ExecutionStateManager {
     this._browserToModuleMap.set(browserId, browserModuleList);
   }
 
+  /**
+   * Pushes the module detail into the moduleMetadata array
+   *
+   * @param {*} metaData
+   */
+  addMetadataToModuleMetadata(metaData) {
```
step2yeung: could I suggest `addToModuleMetadata`

lib/utils/testem-events.js
```diff
@@ -157,27 +176,30 @@ class TestemEvents {
...
    */
-  completedBrowsersHandler(browserCount, browserId, ui, loadBalance, fileName, writeExecutionFile) {
+  completedBrowsersHandler(browserCount, launcherId, ui, commands, currentDate) {
     let browsersCompleted = false;
 
-    this.stateManager.incrementCompletedBrowsers(browserId);
+    this.stateManager.incrementCompletedBrowsers(launcherId);
     if (this.stateManager.getCompletedBrowser() === browserCount) {
-      if (writeExecutionFile && loadBalance) {
+      if (commands.get('writeModuleMetadataFile')) {
```
step2yeung: Could this be its own function?

lib/utils/testem-events.js
```diff
@@ -157,27 +176,30 @@ class TestemEvents {
...
+        const moduleDetailFileName = path.join(this.root, `module-metadata-${currentDate}.json`);
+        const sortedModuleMetadata = getSortedModuleMetaData(this.stateManager.getModuleMetadata());
+
+        writeJsonToFile(moduleDetailFileName, sortedModuleMetadata, { spaces: 2 });
+        ui.writeLine(`\nThis execution module details was recorded at ${moduleDetailFileName}`);
+      }
+
+      if (commands.get('writeExecutionFile') && commands.get('loadBalance')) {
```
step2yeung: Could this be its own function?

lib/commands/exam.js
```diff
@@ -234,62 +245,100 @@ module.exports = TestCommand.extend({
    */
   _generateCustomConfigs(commandOptions) {
```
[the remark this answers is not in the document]
choheekim: Yes, totally agree that we need to refactor the file. I will create a separated pr as this pr is getting huge.

lib/commands/exam.js
```diff
@@ -234,62 +245,100 @@ module.exports = TestCommand.extend({
...
-    const testExecutionFileName = `test-execution-${Date.now()}.json`;
-    const browserCount = Object.keys(config.testPage).length;
-    const browserExitHandler = function() {
-      const browserId = getBrowserId(this.launcher.settings.test_page);
+
+    const browserExitHandler = function(failed = false) {
+
+      const launcherId = this.launcher.id;
+      if (!failed && commands.get('loadBalance')) {
+        try {
```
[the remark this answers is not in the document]
choheekim: Yes, I will create a separated pr to refactor as this pr is getting huge.

lib/utils/testem-events.js
choheekim: Refactoring exam.js and testem-events.js will be in a separated pr as the current pr is getting bigger.

lib/utils/testem-events.js
choheekim: Refactoring exam.js and testem-events.js will be in a separated pr as the current pr is getting bigger.

choheekim:
As @step2yeung leave comments, exam.js and testem-events.js are getting complicated. I will address the issue in a separated pr as this pr is getting big.

lib/commands/exam.js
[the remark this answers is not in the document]
step2yeung: sounds good!

addon-test-support/-private/ember-exam-qunit-test-loader.js
```diff
@@ -118,6 +119,17 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
...
   }
 
+  /**
+   * setupModuleMetadataHandler() register QUnit callback to enable generating module metadata file.
+   */
+  setupModuleMetadataHandler() {
+    this._qunit.moduleDone((metadata) => {
+      // testem:module-done-metadata is sent to server to keep track of test module details.
+      // 'metadata' contains module name, total number of assertion ran in the module, and module runtime.
+      this._testem.emit('testem:module-done-metadata', metadata);
```
Turbo87: this line is breaking our test runs right now, saying:

> Cannot read property 'emit' of undefined

are you using a newer `testem` version? I'm not sure where `this._testem` is coming from 🤔

choheekim: this._testem is coming from https://github.com/ember-cli/ember-exam/pull/265/files#diff-bcd6d6c5d835e949677ec46d11a608dbR17.

When instantiating EmberExamQUnitTestLoader a testem object is passed
https://github.com/ember-cli/ember-exam/blob/master/addon-test-support/load.js#L20

BarryThePenguin: I'm seeing a similar thing in two apps after upgrading ember-exam

Running http://localhost:4200/tests

choheekim: @Turbo87 @BarryThePenguin Hi all, the issue with v4.0.3 is that testem object is not available during `ember serve`. The fix was merged: https://github.com/ember-cli/ember-exam/pull/377

please downgrade ember-exam to v4.02 or upgrade to v4.0.4 ( which is a version include the fix), Thanks!

BarryThePenguin: Great, thanks for getting back to us! 😊

PR #266, opened 2019-04-22
Introducing module-path-filter and test-file-path-filter in ember-exam

choheekim:
ember-test provides `filter` and `modules` to allow executing a subset of the whole test suite. The problem of current provided options is that the filtering is processed after loading a test module, which means that the filtering is done in QUnit level after requiring a module. This results requiring all the test modules in the test suite even though a subset of the tests is targeted to execute.

`module-path-filter` and `test-file-path-filter` option provide to allow running a targeted test modules that are matched with a given module path string and it only loads/requires the targeted test modules which could impact reducing test loading time.

### DIFF
--- a/addon-test-support/-private/ember-exam-mocha-test-loader.js
+++ b/addon-test-support/-private/ember-exam-mocha-test-loader.js
@@ -1,5 +1,6 @@
 import getUrlParams from './get-url-params';
 import splitTestModules from './split-test-modules';
+import { filterTestModules} from './filter-test-modules';
 import { TestLoader } from 'ember-mocha/test-loader';
 
 /**
@@ -47,6 +48,8 @@ export default class EmberExamMochaTestLoader extends TestLoader {
    * Loads the test modules depending on the urlParam
    */
   loadModules() {
+    const modulePathFilter = this._urlParams.get('modulePathFilter');
+    const testFilePathFilter = this._urlParams.get('testFilePathFilter');
     let partitions = this._urlParams.get('partition');
     let split = parseInt(this._urlParams.get('split'), 10);
 
@@ -59,6 +62,14 @@ export default class EmberExamMochaTestLoader extends TestLoader {
     }
 
     super.loadModules();
+
+    if (modulePathFilter || testFilePathFilter) {
+      this._testModules = filterTestModules(
+        this._testModules,
+        modulePathFilter,
+        testFilePathFilter
+      );
+    }
 
     this._testModules = splitTestModules(this._testModules, split, partitions);
     this._testModules.forEach((moduleName) => {
--- a/addon-test-support/-private/filter-test-modules.js
+++ b/addon-test-support/-private/filter-test-modules.js
@@ -1,0 +1,92 @@
+// A regular expression to help parsing a string to verify regex.
+const MODULE_PATH_REGEXP = /^(!?)\/(.*)\/(i?)$/;
+const TEST_PATH_REGEX = /\/tests\/(.*?)$/;
+
+/**
+ * Return the matched test.
+ * e.g. if an input is '!/weight/' it returns an array, ['!/weight/', '!', 'weight', ''];
+ *
+ * @param {*} modulePathFilter
+ */
+function getRegexFilter(modulePathFilter) {
+  return MODULE_PATH_REGEXP.exec( modulePathFilter );
+}
+
+/**
+ * Determine if a given module path is matched with module filter with wildcard.
+ * e.g. A given moduleFilter, /tests/integration/*, matches with /tests/integration/foo and /tests/integration/bar
+ *
+ * @param {*} module
+ * @param {*} moduleFilter
+ */
+function wildcardFilter(module, moduleFilter) {
+  // Generate a regular expression to handle wildcard from path filter
+  const moduleFilterRule = ['^.*', moduleFilter.split('*').join('.*'), '$'].join('');
+  return new RegExp(moduleFilterRule).test(module);
+}
+
+/**
+ * Return a list of test modules that contain a given module path string.
+ *
+ * @param {Array<string>} modules
+ * @param {string} moduleFilter
+ */
+function stringFilter(modules, moduleFilter) {
+  return modules.filter( module => module.includes(moduleFilter) || wildcardFilter(module, moduleFilter) );
+}
+
+/**
+ * Return a list of test modules that matches with a given regular expression.
+ *
+ * @param {Array<string>} modules
+ * @param {Array<string>} modulePathRegexFilter
+ */
+function regexFilter(modules, modulePathRegexFilter) {
+  const re = new RegExp(modulePathRegexFilter[2], modulePathRegexFilter[3]);
+  const exclude = modulePathRegexFilter[1];
+
+  return modules.filter( module => !exclude && re.test(module) || exclude && !re.test(module) );
+}
+
+/**
+ * Return a module path that's mapped by a given test file path.
+ *
+ * @param {*} testFilePathFilter
+ */
+function convertFilePathToModulePath(testFilePathFilter) {
+  const filePathWithNoExtension = testFilePathFilter.replace(/\.[^/.]+$/, '');
+  const testFilePathMatch =  TEST_PATH_REGEX.exec( filePathWithNoExtension );
+  if (typeof testFilePathFilter !== 'undefined' && testFilePathMatch !== null) {
+    return testFilePathMatch[0];
+  }
+
+  return filePathWithNoExtension;
+}
+
+/**
+ * Returns a list of test modules that match with the given module path filter or test file path.
+ *
+ * @param {Array<string>} modules
+ * @param {string} modulePathFilter
+ * @param {string} testFilePathFilter
+ */
+function filterTestModules(modules, modulePathFilter, testFilePathFilter) {
+  // Generates an array with module filter value seperated by comma (,).
+  const moduleFilters = (testFilePathFilter || modulePathFilter).split(',').map( value => value.trim());
+
+  return moduleFilters.reduce((result, moduleFilter) => {
+    const modulePath = convertFilePathToModulePath(moduleFilter);
+    const modulePathRegex = getRegexFilter(modulePath);
+
+    if (modulePathRegex) {
+      return result.concat(regexFilter(modules, modulePathRegex));
+    } else {
+      return result.concat(stringFilter(modules, modulePath));
+    }
+  }, []);
+}
+
+export {
+  convertFilePathToModulePath,
+  filterTestModules
+}
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -49,6 +49,18 @@ module.exports = TestCommand.extend({
         'Randomizes your modules and tests while running your test suite.'
     },
     {
+      name: 'module-path-filter',
+      type: [String],
+      aliases: ['mpf'],
+      description: 'Filters the list of modules to only those that matches by module paths, the value accepts either string or regex.'
+    },
+    {
+      name: 'test-file-path-filter',
+      type: [String],
+      aliases: ['tfpf'],
+      description: 'Filters the list of modules to only those that matches by test file paths, the value accepts either string or regex.'
+    },
+    {
       name: 'replay-execution',
       type: String,
       default: false,
@@ -149,6 +161,22 @@ module.exports = TestCommand.extend({
           );
         }
       }
+    }
+
+    if (commandOptions.modulePathFilter) {
+      commandOptions.query = addToQuery(
+        commandOptions.query,
+        'modulePathFilter',
+        commandOptions.modulePathFilter
+      );
+    }
+
+    if (commandOptions.testFilePathFilter) {
+      commandOptions.query = addToQuery(
+        commandOptions.query,
+        'testFilePathFilter',
+        commandOptions.testFilePathFilter
+      );
     }
 
     if (commandOptions.loadBalance) {
--- a/lib/utils/tests-options-validator.js
+++ b/lib/utils/tests-options-validator.js
@@ -132,7 +132,7 @@ module.exports = class TestsOptionsValidator {
     }
 
     // The parallel option accepts a number, which can be 0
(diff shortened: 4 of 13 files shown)


choheekim:
Working on validating and adding tests against the option.

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,18 @@
...
+ * Returns a list of test modules that match with the given module path filter.
+ *
+ * @param {*} modules
+ * @param {*} modulePathFilter
+ */
+export default function filterTestModules(modules, modulePathFilter) {
+  const tests = [];
+
+  modules.forEach((module) => {
+    if (module.includes(modulePathFilter)) {
```
arthirm: Question, Can regex be supported?

choheekim: current implementation doesn't but I will work on supporting regex as well

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Return the matched test.
+ * e.g. if an input is '!/weight/' it returns an array, ['!/weight/', '!', 'weight', ''];
+ *
+ * @param {*} modulePathFilter
+ */
+function getRegexFilter(modulePathFilter) {
+  return /^(!?)\/([\w\W]*)\/(i?$)/.exec( modulePathFilter );
```
stefanpenner: rather then having a function for this, can we instead:

* `const MODULE_PATH_REGEXP = /.../` at the top of the file, so the regexp can be reused?
* use `MODULE_PATH_REGEXP.exec(modulePathFile)` inline? That way, one can clearly see what it does (hiding this small line behind a function appears as unneeded indirection.)

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Return a list of test modules that contain a given module path string.
+ *
+ * @param {*} modules
+ * @param {*} modulePathFilter
+ */
+function stringFilter(tests, modules, modulePathFilter) {
+  modules.forEach((module) => {
```
step2yeung: We can use filter and return the filtered modules, this way we just return a new array with the filtered modules, instead of modifying the state of the tests array.. 
```
function stringFilter(modules, modulePathFilter) {
  return modules.filter( module => module.includes(modulePathFilter) );
}
```
Then we can simplify filterTestModules as well:
```
function filterTestModules(modules, modulePathFilter) {
  const modulePathRegexFilter = getRegexFilter(modulePathFilter);

  if (modulePathRegexFilter) {
    return regexFilter(tests, modules, modulePathRegexFilter);
  } 
  return stringFilter(tests, modules, modulePathFilter);
} 
```

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param {*} tests
+ * @param {*} modules
+ * @param {*} modulePathRegexFilter
+ */
+function regexFilter(tests, modules, modulePathRegexFilter) {
+  const re = new RegExp(modulePathRegexFilter[2], modulePathRegexFilter[3]);
+  const exclude = modulePathRegexFilter[1];
+
+  modules.forEach((module) => {
```
step2yeung: the filter comment above can also apply here

lib/utils/tests-options-validator.js
```diff
@@ -132,7 +132,7 @@ module.exports = class TestsOptionsValidator {
     }
 
     // The parallel option accepts a number, which can be 0
-    if (this.options.parallel !== undefined) {
+    if (typeof this.options.parallel !== 'undefined') {
```
step2yeung: im curious, what motivated this change?

lib/commands/exam.js
```diff
@@ -48,6 +48,12 @@ module.exports = TestCommand.extend({
       description:
         'Randomizes your modules and tests while running your test suite.'
     },
+    {
+      name: 'module-path-filter',
+      type: String,
+      aliases: ['mpf'],
+      description: 'A module path to filter a module(s) to run.'
```
step2yeung: Suggestion: `Filters the list of modules to only those that matches this value.`
Thoughts?

lib/utils/tests-options-validator.js
choheekim: just to unify things.. in other if block we are using `typeof` for other options so make all the same :)

addon-test-support/-private/ember-exam-mocha-test-loader.js
```diff
@@ -60,6 +62,11 @@ export default class EmberExamMochaTestLoader extends TestLoader {
 
     super.loadModules();
 
+    this._testModules = filterTestModules(
```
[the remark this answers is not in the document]
choheekim: it's because.. I missed it to check whether the filter value is available in mocha-test-loader ;(

lib/commands/exam.js
choheekim: What about A string to filter the list of modules to only those matches the string || the value?

addon-test-support/-private/filter-test-modules.js
```diff
@@ -1,3 +1,38 @@
...
+function stringFilter(modules, modulePathFilter) {
+  return modules.filter( module => module.includes(modulePathFilter) );
+}
+
+/**
+ * Return a list of test modules that matches with a given regular expression.
+ *
+ * @param {*} tests
+ * @param {*} modules
+ * @param {*} modulePathRegexFilter
```
step2yeung: what does this param look like? is it possible for it to be undefined or null? cause if so, line 31 and 32 would throw an error

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,51 @@
...
+  return modules.filter( module => !exclude && re.test(module) || exclude && !re.test(module) );
+}
+
+/**
+ * Returns a list of test modules that match with the given module path filter.
+ *
+ * @param {*} modules
+ * @param {*} modulePathFilter
+ */
+export default function filterTestModules(modules, modulePathFilter) {
```
step2yeung: I like how clean this file looks, good job!

lib/utils/tests-options-validator.js
```diff
@@ -220,6 +220,14 @@ module.exports = class TestsOptionsValidator {
     return shouldRandomize;
   }
 
+
+  /**
+   * TODO
```
step2yeung: looks like the TODO still needs some todos :P

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,51 @@
+const MODULE_PATH_REGEXP = /^(!?)\/([\w\W]*)\/(i?$)/;
```
arthirm: Can you add an explanation of what is the expected regex format here?

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Return a list of test modules that contain a given module path string.
+ *
+ * @param {*} modules
+ * @param {*} modulePathFilter
+ */
+function stringFilter(modules, modulePathFilter) {
+  return modules.filter( module => module.includes(modulePathFilter) );
```
arthirm: Why do we need a separate stringFilter? Can regexFilter not handle this as well?

lib/commands/exam.js
```diff
@@ -48,6 +48,12 @@ module.exports = TestCommand.extend({
       description:
         'Randomizes your modules and tests while running your test suite.'
     },
+    {
+      name: 'module-path-filter',
+      type: String,
+      aliases: ['mpf'],
+      description: 'Filters the list of modules to only those that matches this value.'
```
arthirm: Can we say that it accepts regex as well?

arthirm: Can we say that it accepts regex as well?

addon-test-support/-private/filter-test-modules.js
```diff
@@ -0,0 +1,51 @@
+// A regular expression to help parsing a string to verify regex.
+const MODULE_PATH_REGEXP = /^(!?)\/([\w\W]*)\/(i?$)/;
```
stefanpenner: I believe the following RegExp, maybe a better fit: `/^(!?)\/(.*)\/(i?)$/`

* `$` is at the end
* the character class used inside the `/` and `/` is less restrictive, supporting literally anything. Which supports users providing there own character classes, and other such things.

stefanpenner:
LGTM after https://github.com/ember-cli/ember-exam/pull/266#discussion_r282668973 is addressed.

choheekim:
Update:
- Added `test-file-path-filter` to provide ability of filtering test modules by given test file path
- The regex to convert test file path to module path contains positive lookbehind and it seems not to be supported in javascript. (Trying to find alternative one)

choheekim:
New features introduced so far in this pr:
Two options
1. module-path-filter
2. test-file-path-filter

The two options filter test modules by either module path or test file path. Users can either provide module path in a form of string or regex or test file path. For an option `test-file-path` it also supports wildcard, e.g. /tests/unit/*. 

Soon to update to support to allow multiple paths, e.g. `--module-path-filter=weight, filter` will execute test modules that have either `weight` and `filter` in the module path.

choheekim:
> this LGTM, any further considerations or are we ready to move forward?

It's ready to move forward but I'd love to get another approval from @step2yeung. What do you think?

lib/commands/exam.js
```diff
@@ -48,6 +48,18 @@ module.exports = TestCommand.extend({
...
+      name: 'module-path-filter',
+      type: [String],
+      aliases: ['mpf'],
+      description: 'Filters the list of modules to only those that matches this value, the value accepts either string or regex.'
+    },
+    {
+      name: 'test-file-path-filter',
+      type: [String],
+      aliases: ['tfpf'],
+      description: 'TODO: Add description'
```
step2yeung: reminder: add description

tests/unit/mocha/filter-test-modules-test.js
```diff
@@ -0,0 +1,126 @@
+import { convertFilePathToModulePath, filterTestModules } from 'ember-exam/test-support/-private/filter-test-modules';
+import { describe, it, beforeEach, afterEach } from 'mocha';
+import { expect } from 'chai';
+
+describe('Unit | filter-test-modules', () => {
+  describe('convertFilePathToModulePath', () => {
+    it('should return an input string without file extension when the input contains file extension', () => {
+      expect(convertFilePathToModulePath('/tests/integration/foo.js')).to.deep.equal(
```
step2yeung: Since this is string comparison (and not deep object comparison), we can use `to.equal` instead.
see https://www.chaijs.com/api/bdd/#method_equal

tests/unit/mocha/filter-test-modules-test.js
```diff
@@ -0,0 +1,126 @@
...
+        '/tests/integration/foo'
+      );
+    });
+
+    it(`should return an input string without file extension when the input doesn't contain file extension`, () => {
+      expect(convertFilePathToModulePath('/tests/integration/foo')).to.deep.equal(
```
step2yeung: same here

tests/unit/mocha/filter-test-modules-test.js
```diff
@@ -0,0 +1,126 @@
...
+        '/tests/integration/foo'
+      );
+    });
+
+    it('should return an input string after `tests` when the input is a full test file path', () => {
+      expect(convertFilePathToModulePath('dummy/tests/integration/foo.js')).to.deep.equal(
```
step2yeung: same here

tests/unit/qunit/filter-test-modules-test.js
```diff
@@ -0,0 +1,153 @@
...
+import { convertFilePathToModulePath, filterTestModules } from 'ember-exam/test-support/-private/filter-test-modules';
+
+module('Unit | filter-test-modules', function(hooks) {
+  setupTest(hooks);
+
+  module('covertFilePathToModulePath', function(hooks) {
+    setupTest(hooks);
+
+    test('should return an input string without file extension when the input contains file extension', function(assert) {
+      assert.equal('/tests/integration/foo', convertFilePathToModulePath('/tests/integration/foo.js'))
```
step2yeung: equal's params are `actual, expected`, so i think this should be:
`assert.equal( convertFilePathToModulePath('/tests/integration/foo.js'), '/tests/integration/foo')`
https://api.qunitjs.com/assert/equal

also, missing semicolon at the end

tests/unit/qunit/filter-test-modules-test.js
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    test(`should return an input string without file extension when the input doesn't contain file extension`, function(assert) {
+      assert.equal('/tests/integration/foo', convertFilePathToModulePath('/tests/integration/foo'))
```
step2yeung: same here

tests/unit/qunit/filter-test-modules-test.js
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    test('should return an input string after `tests` when the input is a full test file path', function(assert) {
+      assert.equal('/tests/integration/foo', convertFilePathToModulePath('dummy/tests/integration/foo.js'))
```
step2yeung: same here

choheekim:
Working on addressing review comments and updating docs

choheekim:
> This is looking good, can remove the WIP tag? Could we get docs for this as well?

Yes, I'm currently working on docs :)

choheekim:
@stefanpenner @step2yeung Docs has been updated and all comments are addressed :)

stefanpenner:
@choheekim nice, looks like it needs a rebase. Mind doing that, at which point I'm good with us shipping this.

choheekim:
> @choheekim nice, looks like it needs a rebase. Mind doing that, at which point I'm good with us shipping this.

Rebased and CI is passing now :) Ready to be shipped

PR #274, opened 2019-05-10
Allow graceful exit when async iterator fails to get a module.

step2yeung:
Allow graceful exit during load-balance mode if async iterator fails to get a module.

`--load-balance` mode shouldn't cause a test execution to fail. If it cannot get the next test module, allow other browsers to share the load.

If a hard exit is needed, add `emberExamExitOnError=true` to the url.

***TODO:***
- [x] Add tests

### DIFF
--- a/addon-test-support/-private/async-iterator.js
+++ b/addon-test-support/-private/async-iterator.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const iteratorCompleteResponse = { done: true, value: null };
+
 /**
  * A class to iterate a sequencial set of asynchronous events.
  *
@@ -17,6 +19,7 @@ export default class AsyncIterator {
     // Set a timeout value from either url parameter or default timeout value, 15 s.
     this._timeout = options.timeout || 15;
     this._browserId = options.browserId;
+    this._emberExamExitOnError = options.emberExamExitOnError;
 
     testem.on(this._response, this._boundHandleResponse);
   }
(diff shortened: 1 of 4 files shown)


choheekim:
CI's failing in a test that checks total # of tests after adding one test. Shall we increate the number of tests in `ember-exam/node-tests/acceptance/exam-test.js`

step2yeung:
@choheekim good call! Updated

addon-test-support/-private/async-iterator.js
```diff
@@ -88,20 +91,29 @@ export default class AsyncIterator {
...
+      if (this._emberExamExitOnError) {
+        let err = new Error(
+          `EmberExam: Promise timed out after ${
+            this._timeout
+          } s while waiting for response for ${this._request}`
+        );
+        reject(err);
+      } else {
+        // eslint-disable-next-line no-console
+        console.error(`EmberExam: Promise timed out after ${
```
stefanpenner: when printing, can we use `this.ui.writeLine(message, 'ERROR')` that way, all our stdout/stderr goes through the same facade.

addon-test-support/-private/async-iterator.js
```diff
@@ -88,20 +91,29 @@ export default class AsyncIterator {
...
           this._timeout
-        } s while waiting for response for ${this._request}`
-      );
-      reject(err);
+        } s while waiting for response for ${this._request}. Closing browser to exit gracefully.`);
```
stefanpenner: should we include "retrying count/maximum" in the message, so it's clear what is up?

stefanpenner:
> Allow graceful exit during load-balance mode if async iterator fails to get a module.

Mind listing what circumstance will lead to the iterator not being able to get the next module?

addon-test-support/-private/async-iterator.js
step2yeung: is `this.ui()` available on the browser? I was looking at a few projects like ember-qunit or ember-test-helpers and they use console output, so I haven't found an example to follow.

addon-test-support/-private/async-iterator.js
step2yeung: I don't think we added any retry logic, in that case, do you feel the error message need more to clarify what happened?

Side-note: Although the change in testem to reconnect the socket when a socket disconnect, should help with the problem being addressed by this PR.
With that said, I still feel this PR is still beneficial.

step2yeung:
@stefanpenner Thats a good question.
So far, I've only seen one example that causes the iterator to timeout. This is the socket connection with testem disconnecting.
PR #1340 on testem should have fixed that, as you might be aware :)
With that said, it could still be possible for the timeout to expire if the network is slow, or if the testem server is terminated.

addon-test-support/-private/async-iterator.js
stefanpenner: oh shoot, no you are right. That is only in node-code for ember-addon.s

stefanpenner:
@step2yeung CI looks like it's failing. Once green, I think we can move forward.

step2yeung:
@stefanpenner Fixed the failing test. Good to merge!

stefanpenner:
@step2yeung can you handle the release, or should I?

step2yeung:
@stefanpenner released v3.0.2!

PR #300, opened 2019-06-18
Change option names (module-path-filter and test-file-path-filter) and Enable to execute unique test modules after filtering

choheekim:
1. Change option names to simplify:
- module-path-filter to module-path
- test-file-path-filter to file-path
2. Enable to execute unique test modules after filtering.
e.g. ember exam --file-path='foo/tests/unit/foo.js, /tests/unit/foo.js' 
it executes `foo/tests/unit/foo.js` file once

### DIFF
--- a/addon-test-support/-private/ember-exam-mocha-test-loader.js
+++ b/addon-test-support/-private/ember-exam-mocha-test-loader.js
@@ -48,8 +48,8 @@ export default class EmberExamMochaTestLoader extends TestLoader {
    * Loads the test modules depending on the urlParam
    */
   loadModules() {
-    const modulePathFilter = this._urlParams.get('modulePathFilter');
-    const testFilePathFilter = this._urlParams.get('testFilePathFilter');
+    const modulePath = this._urlParams.get('modulePath');
+    const filePath = this._urlParams.get('filePath');
     let partitions = this._urlParams.get('partition');
     let split = parseInt(this._urlParams.get('split'), 10);
 
@@ -63,11 +63,11 @@ export default class EmberExamMochaTestLoader extends TestLoader {
 
     super.loadModules();
 
-    if (modulePathFilter || testFilePathFilter) {
+    if (modulePath || filePath) {
       this._testModules = filterTestModules(
         this._testModules,
-        modulePathFilter,
-        testFilePathFilter
+        modulePath,
+        filePath
       );
     }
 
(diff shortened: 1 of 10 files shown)


addon-test-support/-private/ember-exam-mocha-test-loader.js
```diff
@@ -48,8 +48,8 @@ export default class EmberExamMochaTestLoader extends TestLoader {
    * Loads the test modules depending on the urlParam
    */
   loadModules() {
-    const modulePathFilter = this._urlParams.get('modulePathFilter');
-    const testFilePathFilter = this._urlParams.get('testFilePathFilter');
+    const modulePath = this._urlParams.get('modulePath');
```
stefanpenner: is this a breaking change?

addon-test-support/-private/ember-exam-qunit-test-loader.js
```diff
@@ -55,8 +55,8 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
   loadModules() {
     const loadBalance = this._urlParams.get('loadBalance');
     const browserId = this._urlParams.get('browser');
-    const modulePathFilter = this._urlParams.get('modulePathFilter');
-    const testFilePathFilter = this._urlParams.get('testFilePathFilter');
+    const modulePath = this._urlParams.get('modulePath');
```
stefanpenner: ^

addon-test-support/-private/ember-exam-mocha-test-loader.js
step2yeung: We never released the previous PR into a version. So would that still be consider breaking?

addon-test-support/-private/ember-exam-qunit-test-loader.js
step2yeung: same applies

addon-test-support/-private/ember-exam-mocha-test-loader.js
stefanpenner: ah, nope then no breaking change.

stefanpenner:
:shipit:

step2yeung:
Looks good to me! 
@choheekim confirming, we are good to merge?

choheekim:
As you all approve it’s good to go! Thank you all for the quick reviewing :)

PR #302, opened 2019-06-19
Throw an error when there are no matching tests with a given input by `file-path` and `module-path`

choheekim:
`filter` and `module` provided by ember-cli throws an error when there are no tests matched with the filter input saying `No tests matched the filter`.  This pr is to throw an error when there are no matching tests with a given input by `file-path` and `module-path` as the same behavior as `filter` and `module`.


choheekim:
> Thanks @choheekim LGTM!

choheekim:
> Thanks @choheekim LGTM!

Merged 👍

PR #308, opened 2019-06-25
Update to use node version >= 8

choheekim:
ember-exam CI has been failing due to dependency issue with `ember-code-snippet`. We are using ember-cli-addon-docs has dependency of ember-code-snippet@^2.4.0 in ember-exam and they released new version (2.4.1) with the changes that make the engine incompatible with node version 6.


scalvert:
👍

This'll need to be a major as it's a breaking change, I believe.

stefanpenner:
> This'll need to be a major as it's a breaking change, I believe.

Yes

choheekim:
A pr in `ember-cli-addon-docs` to bump `ember-codes-snippet version` should be merged first to resolve current CI failure issue
https://github.com/ember-learn/ember-cli-addon-docs/pull/389

package.json
```diff
@@ -30,6 +30,7 @@
     "cli-table3": "^0.5.1",
     "debug": "^4.1.0",
     "ember-cli-babel": "^7.8.0",
+    "ember-cli-htmlbars": "^3.0.1",
```
scalvert: Was it necessary to add this to dependencies?

choheekim: no, moved it back to devDependency :)

step2yeung:
Just want to verify, we went from `^8.10.0` to `8.*`. Was there some original reason we needed `^8.10.0`?

step2yeung:
Merged! Next will release a major version

PR #330, opened 2019-08-01
Remove disable-gpu flag

step2yeung:
Chrome v76 has an issue when starting headless with the `--disable-gpu` flag: https://bugs.chromium.org/p/chromium/issues/detail?id=737678

Turns out this flag was needed to run headless Chrome on Windows, but that limitation has since been addressed and can be removed our testem configs.


stefanpenner:
@step2yeung :shipit:

PR #353, opened 2019-08-29
Regenerate yarn.lock to resolve `Error: Could not find module @ember-data/canary-featuresimported from@ember-data/store/-private.` in CI

choheekim:
Regeneratec yarn.lock to resolve `Error: Could not find module @ember-data/canary-featuresimported from@ember-data/store/-private.` in CI

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,41 +2,41 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0":
-  version "7.0.0"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
-  integrity sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5":
+  version "7.5.5"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     "codeclimate-test-reporter": "^0.5.0",
     "ember-cli": "^3.10.1",
     "ember-cli-htmlbars": "^3.0.1",
-    "ember-cli-addon-docs": "^0.6.13",
+    "ember-cli-addon-docs": "^0.6.0",
```
stefanpenner: if this is a rollback two things:

* the `^` infront should allow the version to still drift up... what is the goal of this specific change, what version do we need?
* if an upstream issue exists, is there a corresponding issue?

choheekim: 1. You were right. it doesn't make any changes. I remade changes to use ^0.6.* so that it's more general. 
2. There isn't a corresponding issue that I'm aware of.

I think this pr is to update yarn.lock so that ember-exam uses latest version of ember-cli-addon-docs along with latest version of ember-data. I will change description :)

package.json
```diff
@@ -42,7 +42,7 @@
     "codeclimate-test-reporter": "^0.5.0",
     "ember-cli": "^3.10.1",
     "ember-cli-htmlbars": "^3.0.1",
-    "ember-cli-addon-docs": "^0.6.13",
+    "ember-cli-addon-docs": "^0.6.*",
```
stefanpenner: I think this change is not needed as the `^` in semver allows the 3rd value to drift anyways.

Is is possible the important part of this PR is actually only that the lock file was regenerated?

choheekim: Thank you for pointing out. Made the pr just to update yarn.lock

choheekim:
@stefanpenner @step2yeung Can I have one more around of review? :)

step2yeung:
LGTM!

PR #357, opened 2019-09-10
Improve complete browser book keeping & improve request next module conditions

step2yeung:
This PR fixes two issues:
1) A browser is considered completed when a `after-tests-complete` or `disconnect` is received. Currently, we keep track of the number of completed browser by incrementing a value. This way of keep track of the completed browsers is prone to issues if duplicate events are fired/received.
To help with this, we keep track of the completed browsers with a map using browserId as the key. This allows unique book keeping of each completed browser.

2) The load-balance functionality uses moduleDone hook from QUnit to request the next test file to load. Since QUnit allows nested modules, it's possible for moduleDone to be called while there are more modules from the test file to be executed. This can cause inefficient load-balancing since a browser in this situation is not free yet. 
To help with this, only request the next test file if `this._qunit.config.queue.length` is zero. `this._qunit.config.queue` is the queue of tests to be executed. If it is not zero, it means there are still test modules to be ran before the browser is considered free.

### DIFF
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -268,10 +268,13 @@ module.exports = TestCommand.extend({
     const ui = this.ui;
     const events = config.custom_browser_socket_events || {};
     const testExecutionFileName = `test-execution-${Date.now()}.json`;
+    const browserCount = Object.keys(config.testPage).length;
     const browserExitHandler = function() {
-      const browserCount = Object.keys(config.testPage).length;
+      const browserId = getBrowserId(this.launcher.settings.test_page);
+
       testemEvents.completedBrowsersHandler(
         browserCount,
+        browserId,
         ui,
         commands.get('loadBalance'),
         testExecutionFileName,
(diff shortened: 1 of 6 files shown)


lib/commands/exam.js
```diff
@@ -268,25 +268,26 @@ module.exports = TestCommand.extend({
...
         commands.get('loadBalance'),
         testExecutionFileName,
         commands.get('writeExecutionFile')
       );
     };
     const browserFailureHandler = function() {
-      if (this.finished) {
-        return;
-      } else if (commands.get('writeExecutionFile')) {
+      if (commands.get('writeExecutionFile')) {
```
step2yeung: Removed the check for `this.finish`. This was needed for `should write test-execution json when browser exits` case to fail. 
When a browser exits with timeout, `this.finish` is already set to true. My understanding of the code, was we want to short circuit this function so we don't run `browserExitHandler` because the way it keeps track of browser completion cannot handle duplicates. But with the new implementation, it is not a problem.

PR #377, opened 2019-09-27
Add to validate if testem is available

choheekim:
with `ember serve` window.testem is not available when loading test-loader.
https://github.com/ember-cli/ember-exam/blob/master/addon-test-support/load.js#L20

This pr is to validate if testem is available to emit an event for module-metadata

### DIFF
--- a/addon-test-support/-private/ember-exam-qunit-test-loader.js
+++ b/addon-test-support/-private/ember-exam-qunit-test-loader.js
@@ -126,7 +126,9 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
     this._qunit.moduleDone((metadata) => {
       // testem:module-done-metadata is sent to server to keep track of test module details.
       // 'metadata' contains module name, total number of assertion ran in the module, and module runtime.
-      this._testem.emit('testem:module-done-metadata', metadata);
+      if (typeof this._testem !== 'undefined' && this._testem !== null) {
+        this._testem.emit('testem:module-done-metadata', metadata);
+      }
     });
   }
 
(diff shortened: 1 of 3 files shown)


addon-test-support/-private/ember-exam-qunit-test-loader.js
```diff
@@ -126,7 +126,9 @@ export default class EmberExamQUnitTestLoader extends TestLoader {
     this._qunit.moduleDone((metadata) => {
       // testem:module-done-metadata is sent to server to keep track of test module details.
       // 'metadata' contains module name, total number of assertion ran in the module, and module runtime.
-      this._testem.emit('testem:module-done-metadata', metadata);
+      if (typeof this._testem !== 'undefined' && this._testem !== null) {
```
stefanpenner: When does this occur, and could a test be written?

choheekim: just added a test to test to validate testLoader loads modules when testem is not available :)

tests/unit/qunit/test-loader-test.js
```diff
@@ -60,6 +60,30 @@ test('loads all test modules by default', function(assert) {
   ]);
 });
 
+test('loads all test modules when testem object is not available', function(assert) {
+  const testemObject = this.testem;
```
step2yeung: Another option to name this: `originalTestem`

tests/unit/qunit/test-loader-test.js
```diff
@@ -60,6 +60,30 @@ test('loads all test modules by default', function(assert) {
...
+  this.testem = undefined;
+  const testLoader = new EmberExamTestLoader(
+    this.testem,
```
step2yeung: is there a reason we don't just pass `undefined` here?

choheekim: we could. My initial thought was that it would be more easy to read if we pass this.testem after assign value to undefined. But I'm opened to any opinions!

Issue #378, opened 2019-09-29
TypeError: Cannot read property 'on' of undefined


Turbo87:
I ran `ember install ember-exam` on a relatively simple Ember addon project, then tried out `ember exam` and this seems to crash when trying to run the tests:

```
Randomizing tests with seed: false
Environment: test
cleaning up...
/Users/tbieniek/Code/qunit-console-grouper/node_modules/ember-exam/lib/commands/exam.js:358
        this.process.on('processExit', browserFailureHandler.bind(this));
                     ^

TypeError: Cannot read property 'on' of undefined
    at BrowserTestRunner.events.tests-start (/Users/tbieniek/Code/qunit-console-grouper/node_modules/ember-exam/lib/commands/exam.js:358:22)
    at Socket.emit (events.js:203:15)
    at /Users/tbieniek/Code/qunit-console-grouper/node_modules/socket.io/lib/socket.js:528:12
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

Turbo87:
It seems like the issue was `import start from 'ember-exam/test-support/start';` missing from the `test-helper.js`, but the error message isn't exactly straight forward and could probably mention that this is what might be missing.

MichalBryxi:
My `test-helper.js` looks like:

```js
import Application from '../app';
import config from 'dummy/config/environment';
import start from 'ember-exam/test-support/start';
import { setApplication } from '@ember/test-helpers';

setApplication(Application.create(config.APP));

start();
```

And I'm getting: 

```
node_modules/ember-exam/lib/commands/exam.js:358
        this.process.on('processExit', browserFailureHandler.bind(this));
                     ^

TypeError: Cannot read property 'on' of undefined
    at BrowserTestRunner.events.tests-start (/node_modules/ember-exam/lib/commands/exam.js:358:22)
    at Socket.emit (events.js:194:15)
    at Socket.EventEmitter.emit (domain.js:441:20)
    at node_modules/socket.io/lib/socket.js:528:12
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

step2yeung:
Following up on this:

I believe @choheekim provided a fix for this issue with #377 which landed on v4.0.4. 
@MichalBryxi @Turbo87 would you be able to try out v4.0.4 and let us know if this is still an issue?

choheekim:
@MichalBryxi Do you mind sharing what ember-exam version you are using?

[...]

PR #388, opened 2019-10-09
Change 'browserFailureHandler' function name to 'browserTerminationHandler' and Add validate to see if a process is available to register process events.

choheekim:
This pr's changes include
1. Name of 'browserFailureHandler' function to 'browserTerminationHandler'
Currently `browesrFailureHandler` is invoked by a multiple events - `processExit`, `processError`, and `disconnect`. The events are emitted when browsers are terminated. However, browser termination doesn't indicate that a browser's exiting due to an error. The name of the function gives a confusion by providing an assumption of the events are invoked by browser errors. 

2. Add validate to see if a process is available to register process events.
`this.process` is only available when browsers are launched via testem. When users manually launch browsers `this.process` is not valid. Due to the reason, checking the process is available before registering process events is required. 

3. Remove `typeof this.process === 'undefined' || this.process === null` in disconnect.
The statement is no longer required to be checked because previously `browserExitHandler` checks browser's completion by recording browser id. Currently implementation checks launcher id rather than browser id which is available for both using `no-launch` and without `no-launch` flag.

### DIFF
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -357,7 +357,12 @@ module.exports = TestCommand.extend({
 
     events['tests-start'] = function() {
       if (!init) {
-        // when test runs without launcher (with --no-launch) process object is not avaialble.
+        // process object is instantiated only when browsers are launched by testem server.
+        // 1. `ember test/exam` where browsers are instantiated by testem - process is available
+        // 2. `ember test/exam --server` where browsers can be instantiated by testem or manually
+        // - process is available only when browsers are instantiated by testem
+        // 3. `ember test/exam --serve --no-launch` where browsers are instantiated manually - process is undefined
+        // 4. `ember serve` where browsers are instantiated manually by developer - process is available.
         if (typeof this.process === 'undefined' || this.process === null) {
           this.process.on('processExit', browserTerminationHandler.bind(this));
           this.process.on('processError', browserTerminationHandler.bind(this));
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -319,8 +319,8 @@ module.exports = TestCommand.extend({
       );
     };
 
-    const browserFailureHandler = function() {
-      // browserFailureHandler is called for disconnect, processError or processExit events.
+    const browserTerminationHandler = function() {
+      // browserTerminationHandler is called for disconnect, processError or processExit events.
       // disconnect and processExit events is fired during global error and successful test runs.
       // On successful test runs, browserExitHandler should already be called. And is unnecessary
       // to call it again, so we should return. This is covered by this.finish = true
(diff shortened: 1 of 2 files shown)


step2yeung:
One thought:
In part of the code where we it is easy to forget all the scenario we have to cover.. it would be good to add some documentation to remind us of all the scenarios we need to cover.
For example, in this case, when we deal with testem and `this.process`, some comment reminding us of all the scenarios:

1. `ember test/exam` where browsers are instantiated by testem
2. `ember test/exam --serve` where browsers can be instantiated by testem or manually
3. `ember test/exam --serve --no-launch` where browsers are instantiated manually
4. `ember serve` where browsers are instantiated manually by developer

choheekim:
> One thought:
> 
> In part of the code where we it is easy to forget all the scenario we have to cover.. it would be good to add some documentation to remind us of all the scenarios we need to cover.
> For example, in this case, when we deal with testem and `this.process`, some comment reminding us of all the scenarios:
> 
> 1. `ember test/exam` where browsers are instantiated by testem
> 2. `ember test/exam --serve` where browsers can be instantiated by testem or manually
> 3. `ember test/exam --serve --no-launch` where browsers are instantiated manually
> 4. `ember serve` where browsers are instantiated manually by developer

choheekim:
@stefanpenner @step2yeung Could you be able to review the pr when you are free? :)

lib/commands/exam.js
```diff
@@ -332,43 +332,45 @@ module.exports = TestCommand.extend({
...
+  _getModuleMetadataAndBrowserExitSocketEvents(browserExitHandler, browserTerminationHandler) {
     const events = {};
     let init = false;
 
     events['tests-start'] = function() {
       if (!init) {
-        this.process.on('processExit', browserFailureHandler.bind(this));
-        this.process.on('processError', browserFailureHandler.bind(this));
+        // when test runs without launcher (with --no-launch) process object is not avaialble.
+        if (typeof this.process === 'undefined' || this.process === null) {
```
step2yeung: @choheekim I remember when you were debugging an issue, this line of code was of interest and may have been a contributor to the issue. What was the outcome of the investigation again?

choheekim: The line was an issue in `disconnect` event where a browser is disconnect and a process is still up a browser is not able to fall under if statement to execute `browserFailureHandler`. 

However, verifying whether process is available in `tests-start` is not required as when a browser is launched manually process is not available then registering callback handler for tests-start will be failing - https://github.com/ember-cli/ember-exam/issues/378

step2yeung: @choheekim is `tests-start is not required` a typo?

choheekim: Woops, it was a typo. Edit the comment :)

PR #389, opened 2019-10-11
[WIP] Add to check qunit config depth

choheekim:


step2yeung:
Curious, what does this PR fix?

choheekim:
> Curious, what does this PR fix?

thought this will fix moduleDone's called when there are multiple top level modules/nested modules but I don't think it's related anymore. Closing the pr

Issue #4, opened 2016-01-06
--split filters out non-test files in tests/


jamesarosen:
I have files in tests like `tests/fixtures/blog-posts.js`. When I use `--split=4 --split-file=<1,2,3,4>`, they get filtered out of some runs, causing test failures.

trentmwillis:
[This is the culprit](https://github.com/trentmwillis/ember-exam/blob/master/lib/utils/tests-processor.js#L13-L15). The intent is that test helpers should be included in all bundles so I filter them out, but this means that that anything that is not a test helper gets broken up as if it is a test. Obviously this is incorrect.

We can reverse the logic to assume anything that is not a test should be included in all bundles. A PR would be welcome, otherwise I will try to get to this sometime this week. (Also need to write tests for this at some point.)

trentmwillis:
Released v0.1.1, should be fixed.

jamesarosen:
:tada:

PR #409, opened 2019-11-05
Check process object is available before register callbacks

choheekim:
This pr is to ensure `process` is available before registering callbacks.

### DIFF
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -363,7 +363,7 @@ module.exports = TestCommand.extend({
         // - process is available only when browsers are instantiated by testem
         // 3. `ember test/exam --serve --no-launch` where browsers are instantiated manually - process is undefined
         // 4. `ember serve` where browsers are instantiated manually by developer - process is available.
-        if (typeof this.process === 'undefined' || this.process === null) {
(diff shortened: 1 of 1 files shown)


gmurphey:
Ran into an issue where the testem runner would crash on `on` not being a method on `this.process`. The change in this PR fixed it! 🎉

step2yeung:
@choheekim anyway to test this?

gmurphey:
@step2yeung To reproduce:

1. create a new app
2. install and setup ember-exam (had to use `ember-cli/ember-exam#4.0.5` as the latest version doesn't seem to be published?)
3. in testem, remove Chrome from the `launch_in_dev` array
4. Run `ember exam --server` and try to open the tests in a browser
5. Observe that the CLI has crashed

choheekim:
@step2yeung I'm not so sure how tests should be like to test this. If you have any suggestions I'd love to hear :)

step2yeung:
@choheekim If you were able to repro and fix locally, I think that's good enough for now. I am thinking if we have tests that does execSync( `ember exam --server`) then validates everything is good. But if not, lets merge this first and explore if we can add --server tests

PR #410, opened 2019-11-05
Fail if parallel is not a numeric value

### LINKED ISSUE #408: --parallel appears to only work when it's the last arg
I ran into an issue with how the args are consumed

`ember exam --split 25 --reporter dot --parallel` will run the entire suite of tests but

`ember exam --split 25 --parallel --reporter dot` will only run the first partition

This is an issue if you've got your test command as `ember exam --split 25 --parallel` and want to change the reporters through the cli

step2yeung:
Fail if parallel is not a numeric value. This fixes #408

### DIFF
--- a/lib/utils/tests-options-validator.js
+++ b/lib/utils/tests-options-validator.js
@@ -258,6 +258,15 @@ module.exports = class TestsOptionsValidator {
    * @return {boolean}
    */
   validateParallel() {
+    const parallelValue = parseInt(this.options.parallel, 10);
+
+    if (isNaN(parallelValue)) {
+      throw new SilentError(
+        `EmberExam: You must specify a Numeric value to 'parallel'. Value passed: ${this.options.parallel}`
+      );
+    }
+    this.options.parallel = parallelValue;
+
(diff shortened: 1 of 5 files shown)


choheekim:
LGTM! Thank you :)

lib/utils/tests-options-validator.js
```diff
@@ -258,6 +258,14 @@ module.exports = class TestsOptionsValidator {
    * @return {boolean}
    */
   validateParallel() {
+    const parallelValue = parseInt(this.options.parallel);
```
rwjblue: Generally speaking, you should always pass a radix to `parseInt` (Note the radix does **not** default to 10!!!! [see MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt)).

```suggestion
    const parallelValue = parseInt(this.options.parallel, 10);
```

step2yeung: Thanks @rwjblue !! Fixed!

choheekim:
CI will be passing once rebase :) !

PR #422, opened 2019-11-13
Update yarn.lock to resolve failure in CI

choheekim:
CI has been having failure due to an issue saying
`/home/travis/build/ember-cli/ember-exam/-private/system/core-store.ts: Imported IDENTIFIERS from @ember-data/canary-features which is not a supported flag.`

this issue is related to https://github.com/emberjs/data/issues/6567 and refreshing yarn.lock resolve the issue.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,18 +10,18 @@
     "@babel/highlight" "^7.0.0"
 
 "@babel/core@^7.0.0", "@babel/core@^7.1.6", "@babel/core@^7.2.2", "@babel/core@^7.3.3", "@babel/core@^7.3.4", "@babel/core@^7.5.5":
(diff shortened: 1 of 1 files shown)


step2yeung:
Thanks for the fix Chohee! Is there a way to track this so that we know when we we can use `^` again on the versioning so we aren't pinned to the v1.6 ?

choheekim:
> Thanks for the fix Chohee! Is there a way to track this so that we know when we we can use `^` again on the versioning so we aren't pinned to the v1.6 ?

Hey @step2yeung! looks like it's an issue with caching, not version of ember-template-lint causing the issue. I've updated this pr to rewrite yarn.lock 
https://github.com/emberjs/data/issues/6567

PR #463, opened 2019-12-24
updating header comments to fix warnings during "ember build"

davecombs:
Doing an 'ember build' for ember-exam generates a large number of yuidoc warning messages, usually for a missing type. This PR is to add the relevant yuidoc entries to fix those warnings during build.

### DIFF
--- a/addon-test-support/-private/async-iterator.js
+++ b/addon-test-support/-private/async-iterator.js
@@ -25,12 +25,19 @@ export default class AsyncIterator {
   }
 
   /**
-   * Return whether the response queue is done.
+   * Indicates whether the response queue is done or not.
+   *
+   * @method done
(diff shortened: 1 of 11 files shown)


step2yeung:
Thanks for the PR @davecombs!
I see `@method` being added to all the functions, which seems a bit redundant to me. Is this annotation truly necessary?

davecombs:
> Thanks for the PR @davecombs!
> I see `@method` being added to all the functions, which seems a bit redundant to me. Is this annotation truly necessary?

I was getting build warnings in the console before doing that. Since that was the purpose of the PR, yes, I'd say it was necessary. :-)

step2yeung:
@davecombs Gotcha, since I see you mention that `usually for a missing type`. Merging PR.

PR #473, opened 2020-01-10
Dropping support for node 8

choheekim:
As node 8 is no longer in maintenance moving node engine in ember-exam from 8 to 10.
https://nodejs.org/en/about/releases/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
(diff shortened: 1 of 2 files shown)


choheekim:
@stefanpenner @step2yeung  Could you guys review the pr? Thank you!

step2yeung:
LGTM, 👋  Node8!

PR #479, opened 2020-01-18
Handle an error thrown in getBrowserId() and Enhance completeBrowserHandler()

choheekim:
**Background**
`--wmmf` and `--wef` are two options in order to generate test execution data such as test module metadata and test execution details, e.g. with `--wmmf` it creates a json file which contains duration, test stats such as number of passed, failed, and skipped tests of modules executed. Current implementation achieves this feature by saving test module information while executing tests and creating a json file with the information as a browser(s) complete running all tests. 

**Problem Statement**
An issue is raised when executing tests with multiple tests. With multiple browsers, ember-exam currently uses a number of browsers passed by `--parallel` with `--load-balance` or `--split` and compares the number to browsers exited. However, if browsers are never launched by some other reasons then exiting browsers never match with the total number of browser(s) that are initially passed. 

For example, when executing `ember exam --load-balance --parallel 10` for some reasons 2 out of 10 browsers were not able to be launched and only 8 browsers served to run tests. A number of browser(s) to be wanted to launch is 10 but exited browsers is 8. Because the two numbers (8 and 10) are different `--wmmf` and `--wef` will not create any files.

**Solution**
The pr changes to compare a number of exited browsers to "launched" browsers rather than number passed by the option (`--split` or `--parallel`). 

Other than the issue,
the pr also makes change the way it handles an error thrown in `getBrowsersId()`.

### DIFF
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -288,16 +288,18 @@ module.exports = TestCommand.extend({
       const launcherId = this.launcher.id;
       if (!failed && commands.get('loadBalance')) {
         try {
-          const browserId = getBrowserId(this.launcher.settings.test_page);
+          const browserId = getBrowserId(this.launcher);
           log.info(`Browser ${browserId} exiting. [ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]`);
         } catch (err) {
-          const message = testemEvents.stateManager.getTestModuleQueue() === null ?
+          const moduleQueueMessage = testemEvents.stateManager.getTestModuleQueue() === null ?
           'testModuleQueue is not set.' :
           `[ # of modules in current module queue ${testemEvents.stateManager.getTestModuleQueue().length} ]`;
+
           if (typeof err === 'object' && err !== null) {
-            err.message = `${err.message} \n ${message}`;
+            err.message = `${err.message} \n ${moduleQueueMessage}`;
+            ui.writeLine(err.message);
           } else {
-            throw new Error(message);
+            throw new Error(moduleQueueMessage);
           }
         }
       }
@@ -334,9 +336,7 @@ module.exports = TestCommand.extend({
         return;
       }
       if (commands.get('writeExecutionFile')) {
-        // when a browser fails with an error record failed browser id.
-        const browserId = getBrowserId(this.launcher.settings.test_page);
-        testemEvents.recordFailedBrowserId(browserId);
+        testemEvents.recordFailedBrowserId(this.launcher, ui);
       }
 
       browserExitHandler.call(this, true);
@@ -353,6 +353,7 @@ module.exports = TestCommand.extend({
    */
   _getModuleMetadataAndBrowserExitSocketEvents(browserExitHandler, browserTerminationHandler) {
     const events = {};
+    const testemEvents = this.testemEvents;
     let init = false;
 
     events['tests-start'] = function() {
@@ -368,6 +369,10 @@ module.exports = TestCommand.extend({
           this.process.on('processError', browserTerminationHandler.bind(this));
         }
         init = true;
+      }
+
+      if (typeof this.launcher !== 'undefined' && this.launcher !== null) {
+        testemEvents.recordStartedLauncherId(this.launcher.id);
       }
     };
 
@@ -407,6 +412,8 @@ module.exports = TestCommand.extend({
    */
   _getLoadBalancingBrowserSocketEvents({ isLoadBalance, isReplayExecution, isWriteExecutionFile } , testemEvents) {
     const events = {};
+    const ui = this.ui;
+
     events['testem:set-modules-queue'] = function(modules, browserId) {
       testemEvents.setModuleQueue(
         browserId,
@@ -424,8 +431,7 @@ module.exports = TestCommand.extend({
     };
     events['test-result'] = function(result) {
       if (result.failed && isWriteExecutionFile) {
-        const browserId = getBrowserId(this.launcher.settings.test_page);
-        testemEvents.recordFailedBrowserId(browserId);
+          testemEvents.recordFailedBrowserId(this.launcher, ui);
       }
     };
 
--- a/lib/utils/test-page-helper.js
+++ b/lib/utils/test-page-helper.js
@@ -123,18 +123,22 @@ function combineOptionValueIntoArray(optionValue) {
 }
 
 /**
- * Returns the browserId within the testPage string
+ * Returns the browserId of launcher
  *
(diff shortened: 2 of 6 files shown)


choheekim:
@dcombslinkedin @stefanpenner @step2yeung Feel free to review the pr :) !

lib/utils/test-page-helper.js
```diff
@@ -123,18 +123,22 @@ function combineOptionValueIntoArray(optionValue) {
...
 
-  if (Array.isArray(browserId) !== null && browserId !== null) {
-    return browserId[1];
-  } else {
-    throw new Error('Browser Id can\'t be found.');
+    if (Array.isArray(browserIdMatch) !== null && browserIdMatch !== null) {
+      return browserIdMatch[1];
+    }
+  } catch(err) {
+    const errMsg = `${err.message} \n${err.stack} \nLauncher Settings: ${JSON.stringify(launcher.settings, null, 2)}`;
```
step2yeung: I like that the error message provides the settings that was generated the error!!

lib/commands/exam.js
```diff
@@ -288,7 +288,7 @@ module.exports = TestCommand.extend({
       const launcherId = this.launcher.id;
       if (!failed && commands.get('loadBalance')) {
         try {
-          const browserId = getBrowserId(this.launcher.settings.test_page);
+          const browserId = getBrowserId(this.launcher);
```
step2yeung: should we also catch and write the error message here?

lib/commands/exam.js
```diff
@@ -424,8 +439,15 @@ module.exports = TestCommand.extend({
...
-        testemEvents.recordFailedBrowserId(browserId);
+        let browserId;
+        try {
+          browserId = getBrowserId(this.launcher);
+        } catch(err) {
+          ui.writeLine(err.message);
+        }
+        if (browserId !== null || typeof browserId !== 'undefined') {
+          testemEvents.recordFailedBrowserId(browserId);
+        }
```
step2yeung: worth considering: abstract this function out for reuse?

lib/commands/exam.js
```diff
@@ -369,6 +378,10 @@ module.exports = TestCommand.extend({
         }
         init = true;
       }
+
+      if (typeof this.launcher !== 'undefined' && this.launcher !== null) {
+        testemEvents.recordStartedLauncherId(this.launcher.id);
+      }
```
step2yeung: should this be within the `init` if block??
since `tests-start` gets called on every time a single test is started.. this path could be executed many times, and many of the times it could be duplicates

lib/utils/testem-events.js
```diff
@@ -202,6 +212,8 @@ class TestemEvents {
         ui.writeLine(`\nExecution was recorded at ${testExecutionPath}`);
       }
 
+      ui.writeLine(`${browsersStarted.size} out of ${browserCount} browser(s) has ran tests.`);
```
step2yeung: is this also recorded in the metadata file? it maybe useful information... 
i'd also consider re-wording the message to clarify `browserStarted.size` is the number of browsers that was launched.

choheekim: The # of browsers launched is not recored in the module metadata file. I thought that module metadata file should only be containing module related information. What about your thoughts?

davecombs: I think both the #requested and the #actually started would be very useful information. Knowing them would explain odd behaviors.

lib/commands/exam.js
choheekim: that's very true, but if  it's within `init` it will only record the first browser that emits `tests-start` and other browsers won't be able to reach to the line to record their launch ids.

step2yeung: Makes sense, feel free to resolve :)

choheekim:
This pr has been tested in ember application to verify whether module metadata file is generated when there is browser(s) that couldn't be launched.

lib/utils/testem-events.js
choheekim: @dcombslinkedin @step2yeung yes I agree that the data is very useful :) do you guys think that the data should be included in either test-execution-file or test-module-metadata? or it's fine to log it out? I will merge this pr after resolving this :)

davecombs: I'd personally like it in the metadata file. Logging it out doesn't help for things like CI runs.

step2yeung: +1 on the metadata file, since I separate the usages of the two files as:
 - test-execution file for functionally rerunning the execution
 - test-module-metadata for informational data that helps with investigations

choheekim: Thank you for the feedback! I've made changes to add the data to the metadata file.
```
{
  "requested": "1 browser(s)",
  "launched": "1 browser(s)",
  "modules": [
    {
      ...
    },
    {
```

davecombs: awesome! Thanks, Chohee!

lib/utils/test-page-helper.js
step2yeung: Reopened this, since I was unsure if this was resolved. @choheekim could you comment on this?

node-tests/unit/utils/test-page-helper-test.js
```diff
@@ -39,13 +39,21 @@ describe('TestPageHelper', function() {
...
-    it('should return 0 if no browser param is found in the test page', function() {
+    it('should throw an error if the launcher does not have test page set', function() {
+      const launcher = {
+        foo: 'bar'
+      }
       assert.throws(() => {
-        getBrowserId('loadBalance');
-      }, /Browser Id can't be found./);
+        getBrowserId(launcher);
+      }, /Launcher Settings:/);
```
step2yeung: This was the test I was referring to for the error message, since i don't actually see a message indicating a problem.

lib/utils/test-page-helper.js
choheekim: the test’s checking whether it contains a given regex. As the error message can be vary I thought it’d be better to test against some message that will always be there.

step2yeung: Got it. That makes sense. So if we disregard the test, were you able to test this path such that `err.message` provides a meaningful message?

choheekim: Yes, I was able to test it manually. 
e.g. the error message would look something like below:
```
Cannot read property 'test_page' of undefined
TypeError: Cannot read property 'test_page' of undefined
    at getBrowserId (/Users/cokim/WODS/ember-exam/lib/utils/test-page-helper.js:45:148)
   ...
Launcher Settings: undefined
```

PR #489, opened 2020-02-01
Fix invalid ES module usage

ef4:
This was not actually valid Javascript. `./-private/patch-testem-output` exports two functions, not a default export. It accidentally works due to an old compatibility feature. It will not work in Embroider or any other build system that follows the ES module spec.


choheekim:
LGTM! Thank you!

PR #612, opened 2020-10-08
Update README.md

jrowlingson:
https://github.com/ember-cli/ember-exam/blob/e146714dd9bb907ee0f83094fa53569cddd1f73f/lib/utils/tests-options-validator.js#L333


rwjblue:
Thank you!

PR #618, opened 2020-10-28
Swap to GitHub actions for CI.

rwjblue:


rwjblue:
Landing this (these failures are the same as the ones on master)

PR #651, opened 2021-02-04
Fix #634: Could not find module `qunit`

### LINKED ISSUE #634: Could not find module `qunit` imported from `(require)`
In https://github.com/rust-lang/crates.io/pull/3080 we are trying to upgrade crates.io to use ember-qunit v5, but it appears like ember-exam might be incompatible currently. It seems to work fine when running `ember test` regularly, but when `ember test` is run via embroider (`USE_EMBROIDER=true`) then the test runner never progresses. When doing the same with `ember serve` and visiting `/tests` the following error appears on the console:

```
Uncaught Error: Could not find module `qunit` imported from `(require)`
    at missingModule (loader.js:247)
    at findModule (loader.js:258)
    at requireModule (loader.js:24)
    at eval (qunit.js:1)
    at Object.../externals/qunit.js (chunk.9a41034696370d422cb6.js:187)
    at __webpack_require__ (chunk.9a41034696370d422cb6.js:80)
    at eval (ember-exam-qunit-test-loader.js:9)
    at Module../node_modules/ember-exam/test-support/-private/ember-exam-qunit-test-loader.js (chunk.4592e8d4fc35b4d1b9b8.js:740)
    at __webpack_require__ (chunk.9a41034696370d422cb6.js:80)
    at getTestLoader (get-test-loader.js:13)
```

/cc @rwjblue

thoov:
When parent applications are using ember-qunit > 5 this causes an invalid lookup for qunit in requirejs because ember-qunit no longer provides the AMD shim for qunit and instead uses ember-auto-import. This change conditionally uses ember-auto-import if that shim is present or not. Additionally, updating test cases for general embroider compatibility.

Fixes: #634

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "silent-error": "^1.1.1"
   },
   "devDependencies": {
-    "@ember/optional-features": "^1.1.0",
+    "@ember/optional-features": "^2.0.0",
     "@embroider/test-setup": "^0.36.0",
     "babel-eslint": "^10.1.0",
     "codeclimate-test-reporter": "^0.5.0",
@@ -95,6 +95,7 @@
     "testdouble": "^3.16.1"
   },
   "peerDependencies": {
+    "@ember/test-helpers": "*",
     "ember-auto-import": "*",
     "ember-mocha": "*",
     "ember-qunit": "*",
--- a/package.json
+++ b/package.json
@@ -95,7 +95,6 @@
     "testdouble": "^3.16.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -93,15 +95,24 @@
     "testdouble": "^3.16.1"
   },
   "peerDependencies": {
+    "@ember/test-helpers": "*",
+    "ember-auto-import": "*",
```
rwjblue, line 99: I think you need ember-auto-import as a dep (not a devDep + peerDep)

PR #745, opened 2021-10-22
Upgrade eslint-plugin-ember from v8.9.1 to v10.5.8

SergeAstapov:
Just an internal clean up

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -64,7 +64,9 @@ module.exports = {
         node: true,
         mocha: true,
       },
-      rules: {},
+      rules: {
+        'ember/no-test-support-import': 'off',
+      },
     },
     {
       // Test files:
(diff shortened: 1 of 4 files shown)


node-tests/fixtures/test-helper-with-load.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable ember/no-test-support-import */
```
SergeAstapov, line 1: latest `eslint-plugin-ember` complaints about import from `test-support` but this is indeed test related code

tests/dummy/app/components/herp-derp.js
```diff
@@ -1,2 +1,4 @@
+/* eslint-disable ember/no-classic-components, ember/require-tagless-components */
```
SergeAstapov, line 1: linter not happy with this file but this is simply herp derp

PR #747, opened 2021-10-24
Upgrade ember-cli-addon-docs and enable Ember v4 scenarios

SergeAstapov:


SergeAstapov:
this was done in #768

PR #787, opened 2021-12-11
ember-try: Fix mocha scenarios

Turbo87:
Those were setting `devDependencies` instead of `npm.devDependencies` 🙈


Turbo87:
well... this is broken... 🙄

Turbo87:
@thoov it looks like you originally worked on this. could you have a look? otherwise I guess we have to disable the mocha scenarios 😞

PR #840, opened 2022-01-21
Upgrade `@embroider/*` packages to 1.0.0

SergeAstapov:
This supersedes #838 and #822 and makes CI pass.
This should help ecosystem migrate to v1 of `@embroider/*` packages.

### DIFF
--- a/tests/unit/qunit/async-iterator-test.js
+++ b/tests/unit/qunit/async-iterator-test.js
@@ -6,7 +6,7 @@ import {
 } from '@embroider/macros';
 
 if (macroCondition(dependencySatisfies('ember-qunit', '*'))) {
-  let { module, test } = importSync('qunit');
+  let { module, test } = importSync('qunit').default;
 
   module('Unit | Qunit | async-iterator', {
     beforeEach() {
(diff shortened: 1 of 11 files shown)


tests/unit/qunit/async-iterator-test.js
```diff
@@ -6,7 +6,7 @@ import {
 } from '@embroider/macros';
 
 if (macroCondition(dependencySatisfies('ember-qunit', '*'))) {
-  let { module, test } = importSync('qunit');
+  let { module, test } = importSync('qunit').default;
```
SergeAstapov, line 9: I assume previous versions of `@embroider/macro` happily "swallowed" qunit not being an ES module but now we need to be explicit.

I assume this has changed in https://github.com/embroider-build/embroider/pull/1076 ad should be way to move forward in such scenarios.

@ef4 would appreciate you can confirm

ef4: I think you're correct. It was a bug that we allowed this to work before with `.default`.

rwjblue (approved):

Issue #129, opened 2018-05-03
separate partition number from test name


thousand:
Currently the test name is added into the test's name (https://github.com/trentmwillis/ember-exam/blob/e3579e97a43dfa10f959ddec813544f7983fd085/addon-test-support/-private/patch-testem-output.js#L11). This tightly couples the partition to the test itself, and makes it difficult to have reporting based on individual tests regardless of which partition they're executed in.

Ideally the partition would be included as a separate field somewhere, and then ember-exam would include or patch the reporting side to display that in log output instead. This way other testem reporters used in parallel (like the default xunit reporter from testem) can be guaranteed that the test name will be consistent from run-to-run and include the partition number as metadata if needed.

PR #963, opened 2022-07-12
Add preserveTestName CLI flag to remove partition and browser

tasha-urbancic:
Closes: https://github.com/ember-cli/ember-exam/issues/129

I came across the same issue as https://github.com/ember-cli/ember-exam/issues/129 . Specifically, when I increased our parallelization and added `--random` we lost CI insights as the output changes. To increase consistency in the output I have added an optional flag `--preserve-test-name` that removes the dynamic segments from the output. This is a first attempt at solving the issue, so perhaps it isn't the best way. 
 
Because this is is my first time contributing, I am not sure if I am missing something important here. I would love feedback. Thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ The [documentation website](https://ember-cli.github.io/ember-exam/) contains ex
     + [Split Test Parallelization](#split-test-parallelization)
   * [Test Load Balancing](#test-load-balancing)
       - [Test Failure Reproduction](#test-failure-reproduction)
+  * [Preserve Test Name](#preserve-test-name)
 - [Advanced Configuration](#advanced-configuration)
   * [Ember Try & CI Integration](#ember-try--ci-integration)
   * [Test Suite Segmentation](#test-suite-segmentation)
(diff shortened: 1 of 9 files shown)


tniezurawski:
This is a real issue. We have insights enabled on CircleCI and because of splitting and parallelizing tests, we can't benefit from it to the fullest. Here's an example of flaky tests. We know they fail more often but the name of tests are not consistent so the counting doesn't work as it should:

image

That of course affects all analyses, like the slowest tests. We see clear duplicates there:

image

How can I help make it happen? Who's approving and releasing in this project? @rwjblue? @Turbo87? @SergeAstapov?

SergeAstapov (approved):

SergeAstapov:
@tniezurawski can't help with pushing this through. but changes looks good to me! thanks for adding tests, this really helps.

README.md
```diff
@@ -375,6 +376,33 @@ $ ember exam --replay-execution=test-execution-000000.json
...
+
+When using `--split` and/or `--load-balance` the output will look something like:
+
+```bash
+# ember exam --split=2 --partition=1 --parallel=3 --load-balance
+ok 1 Chrome 66.0 - Exam Partition 1 - browser Id 1 - some test
+ok 2 Chrome 66.0 - Exam Partition 1 - browser Id 2 - another test
+ok 3 Chrome 66.0 - Exam Partition 1 - browser Id 3 - some the other test
+```
+However, if you change the amount of parallelization, or randomize accross partitions, the output will change for the same test, which may be an issue if you are tracking test insights over time.
```
tniezurawski, line 389: Small typo:

```suggestion
However, if you change the amount of parallelization, or randomize across partitions, the output will change for the same test, which may be an issue if you are tracking test insights over time. 
```

tniezurawski:
Thanks @SergeAstapov 🙌 and thanks @tasha-urbancic for the change 😍

tniezurawski (approved):
Just tested on my CI and see the proper result in `xunit` report (running on 4 containers x 3 browsers each):

```xml
<testcase classname="Chrome 106.0" name="Acceptance | 404 page: should redirect user to 404 page when using wrong URL" time="0.261"/>
```

🎉

tniezurawski:
By looking at [the releases page](https://github.com/ember-cli/ember-exam/releases), I think @Turbo87 and @rwjblue are maintainers of this package. What can we do to make this change happen?

ef4:
Hi @tasha-urbancic and @tniezurawski, would either or both of you be willing to be added as maintainers to this package?

tniezurawski:
@ef4 Yes, I'd like to 👍

tasha-urbancic:
@ef4 same! 👍

PR #1086, opened 2023-04-20
Check module queue existence when logging

elwayman02:
The current code makes a null check for the module queue before reading its length, but before it does that, it tries to access the length in another line of code. We should consistently make null checks in both places if we expect it's possible the queue could be null. This issue was masked in v6 and earlier, because the first access of `length` was inside a try-catch block. Upgrading to v7/v8 would causes some test suites to fail out when the module queue comes back as null. While this may indicate an issue, it's clear that the intent of the existing null check was to allow ember-exam to keep operating and not throw if the queue is null. It wouldn't make sense to instead throw for a line of code that's only used for the sake of logging, if we're protecting other parts of the code.


elwayman02:
beta/canary builds are likely failing for unrelated reasons

PR #1313, opened 2025-02-24
Use ember-exam with vite

NullVoxPopuli:
Here is a demo https://github.com/NullVoxPopuli/vite-ember-exam-demo

downside: does not work with minimal + type=module.

compat / non-type=module required.

Marked as enhancement, because we added a new export.
Also, would be good to get the updated docs on npm.

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,0 +1,1 @@
+export { default as start } from './start';
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The [documentation website](https://ember-cli.github.io/ember-exam/) contains ex
 Installation is as easy as running:
 
 ```bash
-$ ember install ember-exam
+$ npm install --save-dev ember-exam
 ```
 
 ## How To Use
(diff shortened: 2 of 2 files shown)


addon-test-support/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as start } from './start';
+export { default as startEmberExam } from './start';
```
mansona, line 2: Why are you duplicating the name here? 🤔 I would just pick one, probably `start`

amk221:
Should `qunitStart` be renamed `startQunit` to align with `startEmberExam`, (as opposed to emberExamStart)

addon-test-support/index.js
mansona: I think it's fine having ember exam reflect what we're doing in ember-qunit 👍 I think potentially better 🎉 

Could you remove this and change the other thing to `{start as startEmberExam}` ?

NullVoxPopuli: updated

NullVoxPopuli:
failures also occur on `main` and seem like they're due to a floating dep https://github.com/ember-cli/ember-exam/actions/runs/13639548039/job/38133265025?pr=1323

NullVoxPopuli:
The state of `main` has degraded (we didn't have required checks on CI, so renovate is potentially to blame on the breakage there).

I'm going to merge this anyway, as it's more important the docs are updated, than having CI green.

Meanwhile, I'm trying to figure out what went wrong when on main.

PR #1314, opened 2025-02-24
Convert to pnpm

NullVoxPopuli:


.github/workflows/ci.yml
```diff
@@ -18,13 +18,16 @@ jobs:
...
+          run_install: false
       - uses: actions/setup-node@v4
         with:
           node-version: 18
-          cache: 'yarn'
-      - run: yarn install --frozen-lockfile
-      - run: yarn lint
-      - run: yarn test:ember
+          cache: 'pnpm'
+      - run: pnpm install
```
mansona, line 28:
```suggestion
      - run: pnpm install --frozen-lockfile
```

.github/workflows/ci.yml
```diff
@@ -18,13 +18,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - uses: pnpm/action-setup@v4
+        with:
+          run_install: false
```
mansona, line 23:
```suggestion
```

I don't know that you need this before you install node 🤔

mansona (commented):
just 2 comments on the first job in the workflow, but they apply to all the following ones too

.github/workflows/ci.yml
NullVoxPopuli: this is default in CI https://pnpm.io/cli/install, so the flag isn't needed

.github/workflows/ci.yml
NullVoxPopuli: ah nice -- I guess at some point the default changed -- https://github.com/pnpm/action-setup?tab=readme-ov-file#run_install


default used to be `true` i think -- which was very annoying

.github/workflows/ci.yml
mansona: til 🎉

mansona (approved):

PR #1347, opened 2025-03-13
Update setup example for new qunit requirements

elwayman02:
`setupEmberOnerrorValidation` was added to the Vite setup instructions in #1313 but missed for the normal setup guide. This is required as of ember-qunit v9, I believe.


NullVoxPopuli:
It's not required, _exactly_ but new apps would otherwise have 0 test, and then have failing tests out of the box (qunit requires an least one test)

NullVoxPopuli (approved):

Issue #1387, opened 2025-04-28
Does not split tests into partitions under vite


Techn1x:
I run this
`pnpm vite build --mode test`
followed by this
`pnpm ember exam --path=dist --split=20 --partition=2`

But it always runs _all tests_. Output is no different when I change the partition number
Randomized does seem to work though, when I add `--random="abc"` to the end for example.

Have I misconfigured something? Are others able to get partitioned tests with ember-exam under vite?

NullVoxPopuli:
Fixed in #1430

PR #1430, opened 2025-07-22
Support vite

NullVoxPopuli:
~~Depends on: https://github.com/ember-cli/ember-exam/pull/1431 (else diff will be too big)~~ Edit: merged

Applies some changes from: https://github.com/ember-cli/ember-exam/issues/1387

This might be a breaking change because:
- `start()` has to be `async` now (for vite only, and only when passing import.meta.glob (no eager as `availableModules`) -- this is so we don't load all test modules for every partition - but prior users won't notice.

Existing users of ember-exam (broccoli, webpack) hopefully won't be disrupted, and they don't need to change their `start` to be awaited (and this is true of vite as well -- we don't need to await the start function)

> [!IMPORTANT]
> One weird thing about using the `ember exam` command with vite is that the output folder must not be named anything other than `dist`. `ember-cli` (and the tooling team) agree that this restriction is a bug, and `ember test` (as well!) / `ember exam` should be able to handle `--path`s other than `dist`


Techn1x:
This all makes sense to me.

I also eventually realised I needed to make the fns async, but I neglected to update my comments with the code (didn't want to add more unnecessary detail).

Much better approach than what I had, and looks backwards compatible with pre-vite setups. I couldn't figure out a sensible way to do that. Nice work ✌️

Is the failing test an easy fix?

mansona (approved):
This looks good, I'm happy with the tests so it's at least a viable thing for a release 👍 There might be some edge cases that show up in the ecosystem but I think we should release this and fix-forward 💪

PR #1431, opened 2025-07-23
Upgrade eslint / prettier

NullVoxPopuli:
Because only one major of eslint is allowed in a monorepo, I'm extracting lint upgrades to here

extracted from: https://github.com/ember-cli/ember-exam/pull/1430


eslint.config.mjs
```diff
@@ -0,0 +1,30 @@
+import globals from "globals";
+import { ember } from "ember-eslint";
```
NullVoxPopuli, line 2: this library tracks / supports all the currently active blueprints -- it's way easier to adopt in projects migrating because I don't need to add/upgrade/delete the 6+ dependencies that are required to use eslint in the default blueprint

PR #1515, opened 2026-05-28
Handle transient browser disconnections

### LINKED ISSUE #370: Getting No moduleQueue was set when using load-balance
I've tried to use the `load-balance` option in our app, that's already using exam but got an execution error:

command: 

```
ember exam --parallel=3 --load-balance
```

error:
```
throw new Error('No moduleQueue was set.')
    at TestemEvents.nextModuleResponse (/project/node_modules/ember-exam/lib/utils/testem-events.js:107:13)
```

packages:
```
    "ember-cli": "~3.12.0",
    "ember-exam": "^4.0.2",
    "ember-qunit": "^4.5.1",
```

test-helpers.ts:
```
// @ts-ignore
import Application from '../app';
import config from '../config/environment';
import { setApplication } from '@ember/test-helpers';
import start from 'ember-exam/test-support/start';
import 'qunit-dom';
import 'ember-cli-testdouble-qunit';

setApplication(Application.create(config.APP));

start();
```

I'm unsure how to debug this error.

davidtaylorhq:
Testem fires the `disconnect` event even for temporary drops (e.g. CPU contention) as well as terminal exits. Counting those as browser completions is problematic if all other browsers finish before they reconnect, and causes a 'No moduleQueue was set.' error.

This commit drops the `disconnect` listener and instead hooks `runner.finish()`, which fires once per browser when testem definitively gives up. Unforunately testem doesn't expose an event for this, hence the monkey-patch.

Fixes #370.

### DIFF
--- a/lib/commands/exam.js
+++ b/lib/commands/exam.js
@@ -325,24 +325,16 @@ module.exports = TestCommand.extend({
       );
     };
 
+    // Called for processError, processExit, and the finish() wrapper below.
+    // Flag avoids double-counting.
     const browserTerminationHandler = function () {
-      // browserTerminationHandler is called for disconnect, processError or processExit events.
-      // disconnect and processExit events is fired during global error and successful test runs.
-      // On successful test runs, browserExitHandler should already be called. And is unnecessary
-      // to call it again, so we should return. This is covered by this.finish = true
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
re-triggering ci

NullVoxPopuli (approved):

